PR #12, opened 2024-02-16
chore(ci): Update repo files

Yuan325:


src/langchain_google_alloydb_pg/__init__.py
```diff
@@ -12,8 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from langchain_google_alloydb_pg import version as package_version
```
averikitsch, line 15: Actually let's just use "from .version import __version__"

src/langchain_google_alloydb_pg/__init__.py
```diff
@@ -12,8 +12,12 @@
...
+from langchain_google_alloydb_pg.alloydb_chat_message_history import (
+    AlloyDBChatMessageHistory,
+)
 from langchain_google_alloydb_pg.alloydb_engine import AlloyDBEngine, Column
 from langchain_google_alloydb_pg.alloydb_vectorstore import AlloyDBVectorStore
-from langchain_google_alloydb_pg.alloydb_chat_message_history import AlloyDBChatMessageHistory
 
 __all__ = ["AlloyDBEngine", "Column", "AlloyDBVectorStore", "AlloyDBChatMessageHistory"]
+__version__ = package_version.__version__
```
averikitsch, line 23:
```suggestion
__all__ = ["AlloyDBEngine", "Column", "AlloyDBVectorStore", "AlloyDBChatMessageHistory", __version__]
```

src/langchain_google_alloydb_pg/__init__.py
Yuan325: Should I just do `from langchain_google_alloydb_pg.version import version` to make it consistent with the others instead?

averikitsch (approved):

PR #34, opened 2024-02-24
feat: add verify schema to chat message

averikitsch:


src/langchain_google_alloydb_pg/alloydb_chat_message_history.py
```diff
@@ -15,35 +15,86 @@
...
-        if not results:
-            return []
+    @classmethod
+    async def create(
+        cls,
+        engine: AlloyDBEngine,
+        session_id: str,
+        table_name: str,
+    ):
+        table_schema = await engine._aload_document_table(table_name)
```
jackwotherspoon, line 69: should this be `engine._aload_chat_message_history_table(table_name)` instead of document table?

src/langchain_google_alloydb_pg/alloydb_chat_message_history.py
```diff
@@ -15,35 +15,86 @@
...
+        messages: List[BaseMessage],
+    ):
+        if key != AlloyDBChatMessageHistory.__create_key:
+            raise Exception(
+                "Only create class through 'create' or 'create_sync' methods!"
+            )
         self.engine = engine
         self.session_id = session_id
         self.table_name = table_name
+        self.messages = messages
```
jackwotherspoon, line 60: If `.messages` no longer reads from the database and instead uses this list is there a chance the DB and this list could get out of sync somehow?

src/langchain_google_alloydb_pg/alloydb_chat_message_history.py
averikitsch: Updated to be a generic load table function name

src/langchain_google_alloydb_pg/alloydb_chat_message_history.py
averikitsch: I was a little worried about that too but this also means we don't have to query the database each time we read messages

kurtisvg: This also concerns me... are we only writing to the database then?

kurtisvg (changes requested):

kurtisvg (approved):

PR #109, opened 2024-04-25
chore: test against private IP

averikitsch:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ authors = [
     {name = "Google LLC", email = "googleapis-packages@google.com"}
 ]
 dependencies = [
-    "google-cloud-alloydb-connector[asyncpg]>=1.0.0, <2.0.0",
+    "google-cloud-alloydb-connector[asyncpg]>=0.4.0, <2.0.0",
     "langchain-core>=0.1.1, <1.0.0",
     "langchain-community>=0.0.18, <1.0.0",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -9,7 +9,7 @@ authors = [
     {name = "Google LLC", email = "googleapis-packages@google.com"}
 ]
 dependencies = [
-    "google-cloud-alloydb-connector[asyncpg]>=1.0.0, <2.0.0",
+    "google-cloud-alloydb-connector[asyncpg]>=0.4.0, <2.0.0",
```
jackwotherspoon, line 12: is this a mistake?

jackwotherspoon (changes requested):

pyproject.toml
averikitsch: @jackwotherspoon This was accidentally updated. The library should be able to support from v0.4 to present.

jackwotherspoon: the native support for the IP type as a `str` was only added in `1.0.0` of the AlloyDB Python Connector. Since we removed the conversion from this lib as part of https://github.com/googleapis/langchain-google-alloydb-pg-python/pull/102, Python Connector version 1.0.0 or greater is required.

averikitsch: Now updated.

jackwotherspoon (approved):

PR #139, opened 2024-05-30
feat: Add IVF and ScaNN Indexes

duwenxin99:
Added 2 indexes:
1. IVF from pgvector++
2. SCANN from postgres_ann

Design Doc: https://docs.google.com/document/d/1w_-1Q8hdEEPyzoTmS6sruLmpYlSQPMWfUy5Ds4tIXIU/edit?tab=t.0#heading=h.i007da7g9lwp

### DIFF
--- a/docs/vector_store.ipynb
+++ b/docs/vector_store.ipynb
@@ -235,6 +235,29 @@
   },
   {
    "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### AlloyDBEngine for AlloyDB Omni\n",
+    "To create an `AlloyDBEngine` for AlloyDB Omni, you will need to first create an engine with any PostgreSQL-compatible client or code library, and then turn it into an `AlloyDBEngine` by calling `AlloyDBEngine.from_engine()`. Here is an example connection using `SQLAlchemy` with the `Asyncpg` dialect:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import sqlalchemy\n",
+    "# Replace with your own AlloyDB Omni info\n",
+    "connstring = f\"postgresql+asyncpg://{OMNI_USER}:{OMNI_PASSWORD}@{OMNI_HOST}:5432/{OMNI_DATABASE}\"\n",
+    "async_engine = sqlalchemy.ext.asyncio.create_async_engine(\n",
+    "    connstring, isolation_level=\"AUTOCOMMIT\"\n",
+    ")\n",
+    "omni_engine = AlloyDBEngine.from_engine(async_engine)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "metadata": {
     "id": "D9Xs2qhm6X56"
    },
@@ -245,7 +268,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 34,
+   "execution_count": null,
    "metadata": {
     "id": "avlyHEMn6gzU"
    },
@@ -269,7 +292,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "id": "5utKIdq7KYi5",
    "metadata": {
     "id": "5utKIdq7KYi5"
@@ -282,7 +305,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": null,
    "metadata": {
     "colab": {
      "base_uri": "https://localhost:8080/"
@@ -417,6 +440,27 @@
     "\n",
     "index = IVFFlatIndex()\n",
     "await store.aapply_vector_index(index)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `SCANN` index creation (specific to AlloyDB Omni) requires sufficient maintenance work memory. You need to set the database flag `maintenance_work_mem` by calling `set_maintenance_work_mem` before applying the index."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from langchain_google_alloydb_pg.indexes import SCANNIndex\n",
+    "\n",
+    "VECTOR_SIZE = 768   # Replace with the vector size of your embedding model\n",
+    "index = SCANNIndex()\n",
+    "await omni_store.set_maintenance_work_mem(index.num_leaves, VECTOR_SIZE)\n",
+    "await omni_store.aapply_vector_index(index)"
    ]
   },
   {
--- a/docs/vector_store.ipynb
+++ b/docs/vector_store.ipynb
@@ -248,8 +248,11 @@
    "outputs": [],
    "source": [
     "import sqlalchemy\n",
+    "\n",
     "# Replace with your own AlloyDB Omni info\n",
-    "connstring = f\"postgresql+asyncpg://{OMNI_USER}:{OMNI_PASSWORD}@{OMNI_HOST}:5432/{OMNI_DATABASE}\"\n",
+    "connstring = (\n",
+    "    f\"postgresql+asyncpg://{OMNI_USER}:{OMNI_PASSWORD}@{OMNI_HOST}:5432/{OMNI_DATABASE}\"\n",
+    ")\n",
     "async_engine = sqlalchemy.ext.asyncio.create_async_engine(\n",
     "    connstring, isolation_level=\"AUTOCOMMIT\"\n",
     ")\n",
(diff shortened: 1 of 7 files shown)


tests/test_alloydb_vectorstore.py
```diff
@@ -36,6 +36,7 @@
 ]
 
 embeddings = [embeddings_service.embed_query("foo") for i in range(len(texts))]
+num_leaves = 100
```
averikitsch, line 39: Remove this if it's not needed.

src/langchain_google_alloydb_pg/alloydb_vectorstore.py
```diff
@@ -729,6 +730,12 @@ def max_marginal_relevance_search_with_score_by_vector(
         )
         return self.engine._run_as_sync(coro)
 
+    async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> None:
+        index_memory_required = 50 * num_leaves * vector_size * 4
+        await self.engine._aexecute(
+            f"SET maintenance_work_mem = {index_memory_required};"
```
averikitsch, line 736: We need to confirm this is the right format. I see examples using 'TO' instead of '=' and also setting the measurement unit like 'GB'

tests/test_alloydb_vectorstore_index.py
```diff
@@ -123,13 +167,41 @@ async def test_dropindex(self, vs):
...
+        await vs.aapply_vector_index(index, concurrently=True)
+        assert await vs.is_valid_index(DEFAULT_INDEX_NAME)
+        index = IVFIndex(
+            name="secondindex",
+            distance_strategy=DistanceStrategy.PGVECTOR_INNER_PRODUCT,
+        )
+        await vs.aapply_vector_index(index)
+        assert await vs.is_valid_index("secondindex")
+        await vs.adrop_vector_index("secondindex")
+        await vs.adrop_vector_index()
```
averikitsch, line 193: We shouldn't need both drop methods

tests/test_alloydb_vectorstore_index.py
```diff
@@ -123,13 +167,41 @@ async def test_dropindex(self, vs):
...
+        await omni_vs.aapply_vector_index(index, concurrently=True)
+        assert await omni_vs.is_valid_index(DEFAULT_INDEX_NAME)
+        index = SCANNIndex(
+            name="secondindex",
+            distance_strategy=DistanceStrategy.SCANN_COSINE_DISTANCE,
+        )
+        await omni_vs.aapply_vector_index(index)
+        assert await omni_vs.is_valid_index("secondindex")
+        await omni_vs.adrop_vector_index("secondindex")
+        await omni_vs.adrop_vector_index()
```
averikitsch, line 207: Same here

tests/test_alloydb_vectorstore_index.py
```diff
@@ -123,13 +167,41 @@ async def test_dropindex(self, vs):
...
+            distance_strategy=DistanceStrategy.PGVECTOR_INNER_PRODUCT,
+        )
+        await vs.aapply_vector_index(index)
+        assert await vs.is_valid_index("secondindex")
+        await vs.adrop_vector_index("secondindex")
+        await vs.adrop_vector_index()
+
+    async def test_aapply_postgres_ann_index_scann(self, omni_vs):
+        index = SCANNIndex(distance_strategy=DistanceStrategy.SCANN_EUCLIDEAN)
+        await omni_vs.set_maintenance_work_mem(index.num_leaves, VECTOR_SIZE)
```
averikitsch, line 197: We need to add documentation for this

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -28,12 +28,16 @@ class StrategyMixin:
 class DistanceStrategy(StrategyMixin, enum.Enum):
     """Enumerator of the Distance strategies."""
 
-    EUCLIDEAN = "<->", "l2_distance", "vector_l2_ops"
-    COSINE_DISTANCE = "<=>", "cosine_distance", "vector_cosine_ops"
-    INNER_PRODUCT = "<#>", "ip_distance", "vector_ip_ops"
+    PGVECTOR_EUCLIDEAN = "<->", "l2_distance", "vector_l2_ops"
```
averikitsch, line 31: This design makes it harder for our customers. They shouldn't need to care if its PGVector or Scann, they just need a distance strategy. It also doesn't work with our BaseIndex with default distance strategy.

averikitsch (changes requested):

averikitsch:
Please update the Developer.md to note the resources for the Omni instance

src/langchain_google_alloydb_pg/alloydb_vectorstore.py
duwenxin99: updated with units

tests/test_alloydb_vectorstore_index.py
duwenxin99: If we do not drop the first index, it raises `asyncpg.exceptions.DuplicateTableError: relation "langchainvectorindex" already exists` error because the default index name is the same. We can specify index name when initializing the first index, but then we don't test for initializing without index name.

src/langchain_google_alloydb_pg/indexes.py
duwenxin99: I have updated the distance strategy design to add an attribute `scann_index` so we can keep the original strategy structure.

duwenxin99:
Python 12 integration test is failing for this known issue after CPython update: https://github.com/pydantic/pydantic/pull/9612.

docs/vector_store.ipynb
```diff
@@ -233,6 +233,32 @@
     ")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### AlloyDBEngine for AlloyDB Omni\n",
```
averikitsch, line 240: 👏  This is great!

docs/vector_store.ipynb
```diff
@@ -233,6 +233,32 @@
...
+   "outputs": [],
+   "source": [
+    "import sqlalchemy\n",
+    "\n",
+    "# Replace with your own AlloyDB Omni info\n",
+    "connstring = (\n",
+    "    f\"postgresql+asyncpg://{OMNI_USER}:{OMNI_PASSWORD}@{OMNI_HOST}:5432/{OMNI_DATABASE}\"\n",
+    ")\n",
+    "async_engine = sqlalchemy.ext.asyncio.create_async_engine(\n",
+    "    connstring, isolation_level=\"AUTOCOMMIT\"\n",
```
averikitsch, line 257: Is autocommit needed?

docs/vector_store.ipynb
```diff
@@ -419,6 +445,27 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from langchain_google_alloydb_pg.indexes import ScaNNIndex\n",
+    "\n",
+    "VECTOR_SIZE = 768  # Replace with the vector size of your embedding model\n",
+    "index = ScaNNIndex()\n",
+    "await omni_store.set_maintenance_work_mem(index.num_leaves, VECTOR_SIZE)\n",
+    "await omni_store.aapply_vector_index(index)"
```
averikitsch, line 466: Let's use "store" instead of "omni_store", so they are easy copy/pasteable.

tests/test_alloydb_vectorstore_index.py
averikitsch: Thanks for explanation. Please disregard comment.

src/langchain_google_alloydb_pg/alloydb_vectorstore.py
```diff
@@ -731,9 +731,13 @@ def max_marginal_relevance_search_with_score_by_vector(
         return self.engine._run_as_sync(coro)
 
     async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> None:
-        index_memory_required = 50 * num_leaves * vector_size * 4
+        # Required index memory in MB
+        buffer = 1
```
averikitsch, line 735: Is there a reason for the buffer?

src/langchain_google_alloydb_pg/alloydb_vectorstore.py
```diff
@@ -731,9 +731,13 @@ def max_marginal_relevance_search_with_score_by_vector(
...
+        index_memory_required = (
+            round(50 * num_leaves * vector_size * 4 / 1024 / 1024) + buffer
```
averikitsch, line 737: Can you add a comment that this is converting bytes to MB?

averikitsch (approved):

src/langchain_google_alloydb_pg/alloydb_vectorstore.py
duwenxin99: The memory required for scann index creation is rounded to the nearest MB. I set this buffer so that:
1. When the memory calculated is less than 1MB, it wouldn't output 0.
2. In case there is any other maintenance memory usage during index creation, add 1 MB to be safe.

src/langchain_google_alloydb_pg/alloydb_vectorstore.py
duwenxin99: added

docs/vector_store.ipynb
duwenxin99: updated

docs/vector_store.ipynb
duwenxin99: deleted

PR #172, opened 2024-07-09
feat: Rename VectorStore `embeddings` Field

duwenxin99:
Replace the field name `embedding_service` to `embeddings` for vector store.
Note that since `embeddings` is a read-only property, initialize vectorstore class with `self._embeddings = embeddings` and return `self._embeddings` in property method.


samples/index_tuning_sample/requirements.txt
```diff
@@ -1,3 +1,3 @@
-langchain-google-alloydb-pg==0.4.1
-langchain==0.2.7
-langchain-google-vertexai==0.1.3
\ No newline at end of file
+langchain-google-alloydb-pg
```
averikitsch, line 1: Same comment here as in PR #173

duwenxin99: resolved

PR #173, opened 2024-07-09
feat: Remove `alloydb_` prefix from module names

duwenxin99:
Replace module names:
`alloydb_engine` -> `engine`
`alloydb_loader` -> `loader`
`alloydb_indexes` -> `indexes`
`alloydb_vectorstore` -> `vectorstore`
`alloydb_chat_messsage_history` -> `chat_message_history`


samples/index_tuning_sample/requirements.txt
```diff
@@ -1,3 +1,3 @@
-langchain-google-alloydb-pg==0.4.1
-langchain==0.2.7
-langchain-google-vertexai==0.1.3
\ No newline at end of file
+langchain-google-alloydb-pg
```
averikitsch, line 1: Sorry for the confusion. We want to keep these dependencies pinned to a version but we may need to fix the versions to work together. We could have a requirements under samples that is installed instead of the sample-specific requirements.txt

averikitsch (changes requested):

samples/index_tuning_sample/requirements.txt
duwenxin99: Got it. I have replaced with working dependency versions.

averikitsch (approved):

PR #181, opened 2024-07-11
feat: Add similarity search score threshold select function

duwenxin99:
Implements LangChain `_select_relevance_score_fn` function to support similarity search with score threshold: https://github.com/langchain-ai/langchain/blob/cb95198398a66737f1a7f7e0f63417802e03190c/libs/core/langchain_core/vectorstores/base.py#L611

btw fix a titel size in README.rst

### DIFF
--- a/src/langchain_google_alloydb_pg/vectorstore.py
+++ b/src/langchain_google_alloydb_pg/vectorstore.py
@@ -57,7 +57,7 @@ class AlloyDBVectorStore(VectorStore):
         fetch_k: int = 20,
         lambda_mult: float = 0.5,
         index_query_options: Optional[QueryOptions] = None,
-        relevence_score_fn = Optional[Callable[[float], float]] = None,
+        relevence_score_fn: Optional[Callable[[float], float]] = None,
     ):
         if key != AlloyDBVectorStore.__create_key:
             raise Exception(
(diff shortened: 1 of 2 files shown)


tests/test_alloydb_vectorstore_search.py
```diff
@@ -149,6 +149,14 @@ async def test_asimilarity_search_score(self, vs):
         assert results[0][0] == Document(page_content="foo")
         assert results[0][1] == 0
 
+    async def test_similarity_search_with_relevance_scores_threshold(self, vs):
+        score_threshold = {"score_threshold": 0.9}
```
averikitsch, line 153: Can we add a few tests to show the threshold changing the results?

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         fetch_k: int = 20,
         lambda_mult: float = 0.5,
         index_query_options: Optional[QueryOptions] = None,
+        relevance_score_fn: Optional[Callable[[float], float]] = None,
```
averikitsch, line 60: It's not worth supporting an override in the relevance score

averikitsch (changes requested):

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: deleted

tests/test_alloydb_vectorstore_search.py
duwenxin99: Added tests

averikitsch (approved):

PR #183, opened 2024-07-12
chore: improving doc blocks

vishwarajanand:
Adds documentation to some methods for clarity.

* ainit_document_table
* ainit_vectorstore_table

### DIFF
--- a/src/langchain_google_alloydb_pg/engine.py
+++ b/src/langchain_google_alloydb_pg/engine.py
@@ -368,7 +368,7 @@ class AlloyDBEngine:
     ) -> None:
         """
         Create a table for saving of langchain documents.
-        If table already exists, a TABLE_ALREADY_EXISTS error is thrown.
+        If table already exists, a DuplicateTableError error is thrown.
 
         Args:
             table_name (str): The PgSQL database table name.
--- a/src/langchain_google_alloydb_pg/engine.py
+++ b/src/langchain_google_alloydb_pg/engine.py
@@ -293,7 +293,7 @@ class AlloyDBEngine:
                 Default: "langchain_metadata". Optional.
             id_column (str):  Name of the column to store ids.
                 Default: "langchain_id". Optional,
-            overwrite_existing (bool): Whether to drop the existing table.
+            overwrite_existing (bool): Whether to drop the existing table before insertion.
                 Default: False.
             store_metadata (bool): Whether to store metadata in a JSON column if not specified by `metadata_columns`.
                 Default: True.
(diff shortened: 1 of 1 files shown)


src/langchain_google_alloydb_pg/engine.py
```diff
@@ -276,6 +276,28 @@ async def ainit_vectorstore_table(
...
+                Default: "embedding".
+            metadata_columns (List[Column]): A list of Columns to create for custom
+                metadata. Default: []. Optional.
+            metadata_json_column (str): The column to store extra metadata in JSON format.
+                Default: "langchain_metadata". Optional.
+            id_column (str):  Name of the column to store ids.
+                Default: "langchain_id". Optional,
+            overwrite_existing (bool): Whether to drop the existing table.
+                Default: False.
+            store_metadata (bool): Whether to store metadata in the table.
```
averikitsch, line 298:
```suggestion
            store_metadata (bool): Whether to store metadata in a JSON column if not specified by `metadata_columns`.
```

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -276,6 +276,28 @@ async def ainit_vectorstore_table(
...
+                Default: "page_content".
+            embedding_column (str) : Name of the column to store vector embeddings.
+                Default: "embedding".
+            metadata_columns (List[Column]): A list of Columns to create for custom
+                metadata. Default: []. Optional.
+            metadata_json_column (str): The column to store extra metadata in JSON format.
+                Default: "langchain_metadata". Optional.
+            id_column (str):  Name of the column to store ids.
+                Default: "langchain_id". Optional,
+            overwrite_existing (bool): Whether to drop the existing table.
```
averikitsch, line 296:
```suggestion
            overwrite_existing (bool): Whether to drop the existing table before insertion.
```

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -346,11 +368,16 @@ async def ainit_document_table(
     ) -> None:
         """
         Create a table for saving of langchain documents.
+        If table already exists, a TABLE_ALREADY_EXISTS error is thrown.
```
averikitsch, line 371: Add a "Raises:" field after Args. Is the error formatted like "TABLE_ALREADY_EXIST" or is it an OperationalError with message the table already exists?

averikitsch (changes requested):

src/langchain_google_alloydb_pg/engine.py
vishwarajanand: i will rename this to `DuplicateTableError`, becuase the error is actually like below:
```
sqlalchemy.exc.ProgrammingError: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.DuplicateTableError'>: relation <table_name> already exists
[SQL: CREATE TABLE <full command>;]
```

```suggestion
        If table already exists, a DuplicateTableError error is thrown.
```

vishwarajanand: Added raises as well.

averikitsch (approved):

PR #184, opened 2024-07-12
feat: Auto-generate IDs upon adding embeddings

duwenxin99:
If the user doesn't provide IDs when adding vector store embeddings, auto-generate integer IDs for them.

### DIFF
--- a/src/langchain_google_alloydb_pg/vectorstore.py
+++ b/src/langchain_google_alloydb_pg/vectorstore.py
@@ -221,7 +221,7 @@ class AlloyDBVectorStore(VectorStore):
         **kwargs: Any,
     ) -> List[str]:
         if not ids:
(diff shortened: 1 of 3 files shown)


src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -221,7 +221,7 @@ async def _aadd_embeddings(
         **kwargs: Any,
     ) -> List[str]:
         if not ids:
-            ids = ["NULL" for _ in texts]
+            ids = [str(i) for i, _ in enumerate(texts)]
```
averikitsch, line 224: We need to generate an uuid:
```suggestion
            ids = [str(uuid.uuid4()) for _ in texts]
```

averikitsch (changes requested):

averikitsch (approved):

PR #188, opened 2024-07-17
feat: Support IAM account override

duwenxin99:
Allow user to override default IAM credentials


src/langchain_google_alloydb_pg/engine.py
```diff
@@ -131,6 +131,7 @@ def from_instance(
         user: Optional[str] = None,
         password: Optional[str] = None,
         ip_type: Union[str, IPTypes] = IPTypes.PUBLIC,
+        service_account_email: Optional[str] = None,
```
averikitsch, line 134: I am debating if I like the separate field or if we should re-use "user" like the connector and not require the password. That makes the interface cleaner but then could lead users astray if they need a password for built-in auth vs IAM auth.

duwenxin99: Yeah I think it would mislead the user if we allow them to input a built in auth email without requiring password, which also makes it harder to give out clear error messages like we currently do. Separating the fields also makes it clearer to them what auth method they are using, which is important because user account for built in auth and IAM service account are essentially very different things.

averikitsch (commented):
please add tests

averikitsch (approved):

PR #196, opened 2024-07-29
docs: added vector store initialization from documents

vishwarajanand:
Fixes [b/343215179](http://b/343215179) 🦕


docs/vector_store.ipynb
```diff
@@ -423,6 +423,26 @@
     "print(docs)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Initialize Vector Store with Documents from Any Source\n"
```
vishwarajanand, line 430:
```suggestion
    "### Initialize Vector Store with documents from any source\n"
```

docs/vector_store.ipynb
```diff
@@ -423,6 +423,26 @@
     "print(docs)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Initialize Vector Store with documents from any source\n"
```
averikitsch, line 430: This doesn't make sense just added to the end of this section. It may be best bundled with "Initialize a default AlloyDBVectorStore" with a comment on the difference. The goal is to make it easier and more clear for users that are use to looking at docs like https://python.langchain.com/v0.1/docs/integrations/vectorstores/pinecone/

vishwarajanand: Sure, moved it.

docs/vector_store.ipynb
```diff
@@ -349,6 +349,48 @@
     ")"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Initialize Vector Store with documents from any source\n",
```
averikitsch, line 356: We need to describe the difference between the default above the this option. Let's remove "from any source" because that is potentially confusing to users. We should also add a note that this is a great way to get started but the default way is most likely how an app will use the integration since we won't accidentally add the same documents.

docs/vector_store.ipynb
```diff
@@ -349,6 +349,48 @@
...
+    "\n",
+    "If you already have the documents, create AlloyDBVectorStore with documents already added.\n",
+    "\n",
+    "Each document becomes a row in the table."
```
averikitsch, line 360: Not sure if we need this additional note

docs/vector_store.ipynb
```diff
@@ -349,6 +349,48 @@
...
+    "        page_content=\"Orange Navel\",\n",
+    "        metadata={\"description\": \"orange\", \"content\": \"3\", \"category\": \"fruit\"},\n",
+    "    ),\n",
+    "]\n",
+    "\n",
+    "ids = [str(uuid.uuid4()) for i in range(len(docs))]\n",
+    "\n",
+    "store_with_documents = await AlloyDBVectorStore.afrom_documents(\n",
+    "    docs, embedding, engine, TABLE_NAME, ids=ids\n",
+    ")"
```
averikitsch, line 391: nit: let's use field names like the default example

averikitsch (approved):

PR #203, opened 2024-08-19
fix: Fix `QueryOptions` not applied to similarity search bug

duwenxin99:
The index `QueryOptions` are set by executing `SET LOCAL`, whose effect only last within the same transaction. However, currently the similarity search is in a different transaction, which means the query options are never applied. I fixed this by creating a new `_afetch_with_query_options(query, query_options)` method that executes the query options statement before fetching similarity search result, and modify the search APIs to call this method instead of `_afetch()`

### DIFF
--- a/src/langchain_google_alloydb_pg/engine.py
+++ b/src/langchain_google_alloydb_pg/engine.py
@@ -335,7 +335,7 @@ class AlloyDBEngine:
         """Set temporary database flags and fetch results from a SQL query."""
         async with self._engine.connect() as conn:
             await conn.execute(text(query_options))
-            result = await conn.execute(text(query), params)
+            result = await conn.execute(text(query))
             result_map = result.mappings()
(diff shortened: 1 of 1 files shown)


src/langchain_google_alloydb_pg/engine.py
```diff
@@ -324,11 +324,23 @@ async def _aexecute(self, query: str, params: Optional[dict] = None) -> None:
             await conn.commit()
 
     async def _aexecute_outside_tx(self, query: str) -> None:
-        """Execute a SQL query."""
+        """Execute a SQL query in a new transaction."""
         async with self._engine.connect() as conn:
             await conn.execute(text("COMMIT"))
             await conn.execute(text(query))
 
+    async def _afetch_with_query_options(
```
averikitsch, line 332: nit: do we need to support the params arg here?

averikitsch (approved):

PR #204, opened 2024-08-20
docs: Sample to use AlloyDB interface for vector search in PGVector database

twishabansal:
This code samples shows users how to use AlloyDB interface for vector search in any DB created using the PGVector interface.

This would allow you to migrate from using [PGVector](https://api.python.langchain.com/en/latest/vectorstores/langchain_postgres.vectorstores.PGVector.html#langchain_postgres.vectorstores.PGVector) to [AlloyDB Vector Store](https://github.com/googleapis/langchain-google-alloydb-pg-python/blob/main/docs/vector_store.ipynb) search methods.

### DIFF
--- a/samples/pg_to_alloy_migration_search.ipynb
+++ b/samples/pg_to_alloy_migration_search.ipynb
@@ -317,8 +317,8 @@
     "\n",
     "PGvector sets up two tables. \n",
     "\n",
-    "1. langchain_pg_collection: contains the uuid corresponding to each collection.\n",
-    "2. langchain_pg_embedding: contains documents belonging to all collections."
+    "1. `langchain_pg_collection`: contains the uuid corresponding to each collection.\n",
+    "2. `langchain_pg_embedding`: contains documents belonging to all collections."
    ]
   },
   {
--- a/samples/pg_to_alloy_migration_search.ipynb
+++ b/samples/pg_to_alloy_migration_search.ipynb
@@ -54,7 +54,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 1,
+      "execution_count": null,
       "metadata": {
         "colab": {
           "base_uri": "https://localhost:8080/"
@@ -62,15 +62,7 @@
         "id": "t-FWrCqVH7v5",
         "outputId": "e9e30198-d2ca-4365-b21b-d6ab0eeb0623"
       },
-      "outputs": [
-        {
-          "name": "stdout",
-          "output_type": "stream",
-          "text": [
-            "Note: you may need to restart the kernel to use updated packages.\n"
-          ]
-        }
-      ],
+      "outputs": [],
       "source": [
         "%pip install --upgrade --quiet  langchain-google-alloydb-pg langchain-google-vertexai"
       ]
@@ -86,7 +78,7 @@
     },
     {
       "cell_type": "code",
-      "execution_count": 2,
+      "execution_count": null,
       "metadata": {
         "id": "ziswhCwcY_MV"
       },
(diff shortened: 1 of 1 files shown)


samples/pg_to_alloy_migration_search.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "outputs": [],
+   "source": [
+    "# The PGVector class by default creates a table which contains 4 columns: document, id, cmetadata, collection_id\n",
+    "# Modify column names if you're using different ones instead.\n",
+    "embedding_vectorstore = AlloyDBVectorStore.create_sync(\n",
+    "    engine=engine,\n",
+    "    table_name=\"langchain_pg_embedding\",\n",
+    "    embedding_service=embeddings_service,\n",
+    "    content_column=\"document\",\n",
+    "    metadata_columns=[\"cmetadata\", \"collection_id\"],\n",
```
averikitsch, line 402: cmetadata should be used with metadata_json_column

samples/pg_to_alloy_migration_search.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "id": "iY8oA59vcL7_"
+   },
+   "source": [
+    "### Initialize an AlloyDB Loader\n",
```
averikitsch, line 350: Update this title and description to explain the goal of this section. I initially thought you were loading the docs out of vector store and not just finding the uuid

samples/pg_to_alloy_migration_search.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "phMW5OS1IOA9"
+   },
+   "outputs": [],
+   "source": [
+    "from langchain_google_alloydb_pg import AlloyDBLoader\n",
+    "\n",
+    "# Choose the collection to perform vector search on\n",
```
averikitsch, line 365: Please provide the context on how PGVector is setup so the user can follow the train of thought that we don't want to just search the whole vector store since all of the embeddings are in the same table

samples/pg_to_alloy_migration_search.ipynb
```diff
@@ -0,0 +1,488 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "fE3U6VKRIdst"
+   },
+   "outputs": [],
+   "source": [
+    "# The PGVector class by default creates a table which contains 4 columns: document, id, cmetadata, collection_id\n",
+    "# Modify column names if you're using different ones instead.\n",
+    "embedding_vectorstore = AlloyDBVectorStore.create_sync(\n",
```
averikitsch, line 397: Add a comment that we also support native async functionality

samples/pg_to_alloy_migration_search.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    "colab": {
+     "base_uri": "https://localhost:8080/"
+    },
+    "id": "eVwZdZ35VzZJ",
+    "outputId": "20ef5f10-147e-48f5-efe9-685bb74f0af5"
+   },
+   "outputs": [],
+   "source": [
+    "embedding_vectorstore.similarity_search(\n",
+    "    \"cats\", k=5, filter=f\"collection_id='{uuid}' and cmetadata->>'topic' = 'animals'\"\n",
```
averikitsch, line 460: This is a great example! 👏

samples/pg_to_alloy_migration_search.ipynb
```diff
@@ -0,0 +1,488 @@
...
+    ")"
+   ]
+  }
```
averikitsch, line 463: Do we want to also provide guidance on migrate collections to new tables? This would reduce the need to add a filter on each similarity search. If so can you give more background on why we think they should migrate there tables to individual tables. We will also need to show the users how to create a new table, load the old data and add the documents.

averikitsch (changes requested):

samples/pg_to_alloy_migration_search.ipynb
twishabansal: Thank you!

samples/pg_to_alloy_migration_search.ipynb
twishabansal: I have a Colab Notebook with code that users can reference to migrate to the new tables. The notebook contains several code blocks and SQL commands for data migration.

However, we're also currently developing a migration method within our engine class. It might be better to wait for this implementation and then provide users with a concise code snippet to facilitate their migration process.

What do you think?

averikitsch:
@twishabansal This PR can be close in favor of #249 , correct?

PR #207, opened 2024-08-26
feat: Add multi-modal support vector store

duwenxin99:
1. Add new image search APIs `similarity_search_image()` and `asimilarity_search_image()`.
2. Add `add_images()` and `aadd_images()` endpoints to add images to vector store.
3. Add tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,8 @@ test = [
     "mypy==1.11.2",
     "pytest-asyncio==0.24.0",
     "pytest==8.3.3",
-    "pytest-cov==5.0.0"
+    "pytest-cov==5.0.0",
+    "Pillow==10.4.0"
 ]
 
 [build-system]
--- a/src/langchain_google_alloydb_pg/async_vectorstore.py
+++ b/src/langchain_google_alloydb_pg/async_vectorstore.py
@@ -15,6 +15,7 @@
 # TODO: Remove below import when minimum supported Python version is 3.10
 from __future__ import annotations
 
+import base64
 import json
 import uuid
 from typing import Any, Callable, Iterable, List, Optional, Sequence, Tuple, Type, Union
@@ -306,6 +307,33 @@ class AsyncAlloyDBVectorStore(VectorStore):
         ids = await self.aadd_texts(texts, metadatas=metadatas, ids=ids, **kwargs)
         return ids
 
+    def _encode_image(self, uri: str) -> str:
+        """Get base64 string from image URI."""
+        with open(uri, "rb") as image_file:
+            return base64.b64encode(image_file.read()).decode("utf-8")
+
+    async def aadd_images(
+        self,
+        uris: List[str],
+        metadatas: Optional[List[dict]] = None,
+        ids: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        """Embed images and add to the table."""
+        encoded_images = []
+        if metadatas is None:
+            metadatas = [{"image_uri": uri} for uri in uris]
+
+        for uri in uris:
+            encoded_image = self._encode_image(uri)
+            encoded_images.append(encoded_image)
+
+        embeddings = self._images_embedding_helper(uris)
+        ids = await self._aadd_embeddings(
+            encoded_images, embeddings, metadatas=metadatas, ids=ids, **kwargs
+        )
+        return ids
+
     async def adelete(
         self,
         ids: Optional[List] = None,
@@ -508,6 +536,35 @@ class AsyncAlloyDBVectorStore(VectorStore):
 
         return await self.asimilarity_search_by_vector(
             embedding=embedding, k=k, filter=filter, **kwargs
+        )
+
+    def _images_embedding_helper(self, image_uris: List[str]) -> List[List[float]]:
+        # check if `embed_image()` API is supported by the embedding service used
+        if hasattr(self.embedding_service, "embed_image"):
+            try:
+                embeddings = self.embedding_service.embed_image(image_uris)
+            except Exception as e:
+                raise Exception(
+                    f"Make sure your selected embedding model supports list of image URIs as input. {str(e)}"
+                )
+        else:
+            raise ValueError(
+                "Please use an embedding model that supports image embedding."
+            )
+        return embeddings
+
+    async def asimilarity_search_image(
+        self,
+        image_uri: str,
+        k: Optional[int] = None,
+        filter: Optional[str] = None,
+        **kwargs: Any,
+    ) -> List[Document]:
+        """Return docs selected by similarity search on query."""
+        embedding = self._images_embedding_helper([image_uri])[0]
+
+        return self._engine._run_as_sync(
+            self.__vs.asimilarity_search_by_vector(embedding, k, filter, **kwargs)
         )
 
     def _select_relevance_score_fn(self) -> Callable[[float], float]:
@@ -756,17 +813,6 @@ class AsyncAlloyDBVectorStore(VectorStore):
             results = result_map.fetchall()
         return bool(len(results) == 1)
 
-    def similarity_search(
-        self,
-        query: str,
-        k: Optional[int] = None,
-        filter: Optional[str] = None,
-        **kwargs: Any,
-    ) -> List[Document]:
-        raise NotImplementedError(
-            "Sync methods are not implemented for AsyncAlloyDBVectorStore. Use AlloyDBVectorStore interface instead."
-        )
-
     def add_texts(
         self,
         texts: Iterable[str],
@@ -782,6 +828,17 @@ class AsyncAlloyDBVectorStore(VectorStore):
         self,
         documents: List[Document],
         ids: Optional[List] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBVectorStore. Use AlloyDBVectorStore interface instead."
+        )
+
+    def add_images(
+        self,
+        uris: List[str],
+        metadatas: Optional[List[dict]] = None,
+        ids: Optional[List[str]] = None,
         **kwargs: Any,
     ) -> List[str]:
         raise NotImplementedError(
@@ -834,6 +891,28 @@ class AsyncAlloyDBVectorStore(VectorStore):
         metadata_json_column: str = "langchain_metadata",
         **kwargs: Any,
     ) -> AsyncAlloyDBVectorStore:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBVectorStore. Use AlloyDBVectorStore interface instead."
+        )
+
+    def similarity_search(
+        self,
+        query: str,
+        k: Optional[int] = None,
+        filter: Optional[str] = None,
+        **kwargs: Any,
+    ) -> List[Document]:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBVectorStore. Use AlloyDBVectorStore interface instead."
+        )
+
+    def similarity_search_image(
+        self,
+        image_uri: str,
+        k: Optional[int] = None,
+        filter: Optional[str] = None,
+        **kwargs: Any,
+    ) -> List[Document]:
         raise NotImplementedError(
             "Sync methods are not implemented for AsyncAlloyDBVectorStore. Use AlloyDBVectorStore interface instead."
         )
--- a/src/langchain_google_alloydb_pg/async_vectorstore.py
+++ b/src/langchain_google_alloydb_pg/async_vectorstore.py
@@ -319,7 +319,16 @@ class AsyncAlloyDBVectorStore(VectorStore):
         ids: Optional[List[str]] = None,
         **kwargs: Any,
     ) -> List[str]:
-        """Embed images and add to the table."""
+        """Embed images and add to the table.
+
+        Args:
+            uris (List[str]): List of local image URIs to add to the table.
+            metadatas (Optional[List[dict]]): List of metadatas to add to table records.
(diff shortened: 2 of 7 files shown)


src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -711,17 +882,170 @@ async def asimilarity_search_with_score_by_vector(
...
+                if similarity >= score_threshold
+            ]
+            if len(docs_and_similarities) == 0:
+                warnings.warn(
+                    "No relevant docs were retrieved using the relevance score"
+                    f" threshold {score_threshold}"
+                )
+        return docs_and_similarities
+
+    async def asimilarity_search_with_relevance_scores(
```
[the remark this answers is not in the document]
duwenxin99: These functions don't take image_uri as inputs so it raises error for both the image and text search because the LangChain implementation doesn't specify input parameter name when calling corresponding functions (so it misunderstands kwargs as image_uri input which raised error). This is why I modify it to take image_uri and added parameter specification for this function.

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -613,28 +763,48 @@ async def __query_collection(
...
+        image_uri: Optional[str] = None,
+    ) -> List[float]:
+        if query and image_uri:
+            raise ValueError(
+                f"`query` and `image_uri` cannot be specified at the same time."
+            )
+        elif query:
+            embedding = self.embedding_service.embed_query(text=query)
+        elif image_uri:
+            embedding = self.embedding_service.embed_image([image_uri])[0]
```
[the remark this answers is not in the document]
duwenxin99: Most integrations' `embed_image` API accept a list of input images and returns a list of image embeddings (although the API name is singular). Here we are only embedding 1 input image_uri instead of a list, so we want to return the item in the list by indexing.

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -341,6 +343,39 @@ async def aadd_documents(
...
+        encoded_images = []
+        if metadatas is None:
+            metadatas = [{"image_uri": uri} for uri in images]
+
+        for uri in images:
+            encoded_image = self.encode_image(uri)
+            encoded_images.append(encoded_image)
+
+        if hasattr(self.embedding_service, "embed_image"):
+            embeddings = self.embedding_service.embed_image(images)
```
averikitsch, line 368: We should prepare some documentation for how to use this functionality. Current vertexaiembedding doesn't work since we don't have support for a list

requirements.txt
```diff
@@ -4,3 +4,4 @@ numpy===1.24.4; python_version <= "3.8"
 numpy==1.26.4; python_version > "3.8"
 pgvector==0.3.2
 SQLAlchemy[asyncio]==2.0.34
+Pillow==10.4.0
```
averikitsch, line 7: Where is this required? We may have to add this as a dependency to the library in pyproject.toml

src/langchain_google_alloydb_pg/vectorstore.py
averikitsch: For now we might want to do a try/except to ensure all Embedding service work here until we sort out the VertexAIEmbedding changes

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -349,6 +351,39 @@ async def aadd_documents(
         ids = await self.aadd_texts(texts, metadatas=metadatas, ids=ids, **kwargs)
         return ids
 
+    def encode_image(self, uri: str) -> str:
```
averikitsch, line 354: Make this a private method

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -499,6 +546,57 @@ async def afrom_documents(  # type: ignore[override]
         await vs.aadd_texts(texts, metadatas=metadatas, ids=ids, **kwargs)
         return vs
 
+    @classmethod
+    async def afrom_images(  # type: ignore[override]
```
averikitsch, line 550: I don't love the "from_" methods, so I may have us hold off here until this is enforced by the interface

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -631,28 +782,48 @@ async def __query_collection(
             results = await self.engine._afetch(stmt)
         return results
 
+    def embedding_helper(
```
averikitsch, line 785: Make this a private method

src/langchain_google_alloydb_pg/vectorstore.py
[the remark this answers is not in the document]
averikitsch: Ah, another interface issue. This doesn't work with VertexAIEmbeddings either. 
@kurtisvg is there a better way to try to support services that take either a list or a single string than a try/except statement?

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -349,6 +351,39 @@ async def aadd_documents(
...
+        ids: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        """Embed images and add to the table."""
+        encoded_images = []
+        if metadatas is None:
+            metadatas = [{"image_uri": uri} for uri in images]
+
+        for uri in images:
+            encoded_image = self.encode_image(uri)
```
averikitsch, line 372: Right now we don't support GCS uris since we are dowloading everything, correct?

averikitsch (changes requested):
I manually tested this out and it's looking good. It might be helpful in your doc to describe the milestones of what is completed vs not completed ie GCS support.

Tagging in @kurtisvg as a python devex review on using both query/image_uris as inputs and how to handle differences in the embed_image interfaces while we sort out the interface with LangChain.

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: Right now we support local URI for all embedding services, and GCS URI for Vertex AI embedding because its `embed_image()` implementation accepts GCS URI input: https://github.com/langchain-ai/langchain-google/blob/72d1b5ad853dd1d80c5f74089290caba712221fe/libs/vertexai/langchain_google_vertexai/embeddings.py#L470

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: updated

requirements.txt
duwenxin99: It is required to the testing files only and I have removed it from here to specify in pyproject.toml's test dependencies.

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: removed

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: updated

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: added try/except statement

src/langchain_google_alloydb_pg/vectorstore.py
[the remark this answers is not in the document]
duwenxin99: added try/except statement for now

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -349,6 +351,45 @@ async def aadd_documents(
...
+
+        for uri in images:
+            encoded_image = self._encode_image(uri)
+            encoded_images.append(encoded_image)
+
+        # check if `embed_image()` API is supported by the embedding service used
+        if hasattr(self.embedding_service, "embed_image"):
+            try:
+                embeddings = self.embedding_service.embed_image(images)
+            except Exception as e:
```
averikitsch, line 379: Is there a more specific Exception we can catch? We are also assuming VertexAI will soon support this, right?

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -349,6 +351,45 @@ async def aadd_documents(
...
         return ids
 
+    def _encode_image(self, uri: str) -> str:
+        """Get base64 string from image URI."""
+        with open(uri, "rb") as image_file:
+            return base64.b64encode(image_file.read()).decode("utf-8")
+
+    async def aadd_images(
+        self,
+        images: List[str],
```
averikitsch, line 361: Other implementations use `uris` as this field name

tests/test_alloydb_vectorstore_search.py
```diff
@@ -136,42 +150,99 @@ async def vs_custom(self, engine_sync):
...
         results = await vs.asimilarity_search_with_relevance_scores(
-            "foo", **score_threshold
+            query="foo", image_uri=None, **score_threshold
         )
         assert len(results) == 4
 
         score_threshold = {"score_threshold": 0.02}
         results = await vs.asimilarity_search_with_relevance_scores(
-            "foo", **score_threshold
+            "foo", image_uri=None, **score_threshold
```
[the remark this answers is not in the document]
duwenxin99: Yes it is now required, otherwise the function would take `score_threshold` as `image_uri`.

pyproject.toml
```diff
@@ -44,7 +44,8 @@ test = [
     "mypy==1.11.2",
     "pytest-asyncio==0.24.0",
     "pytest==8.3.2",
-    "pytest-cov==5.0.0"
+    "pytest-cov==5.0.0",
+    "Pillow==10.4.0"
```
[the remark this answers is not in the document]
duwenxin99: Yes correct

tests/test_alloydb_vectorstore_search.py
```diff
@@ -136,42 +150,99 @@ async def vs_custom(self, engine_sync):
...
         assert results == [Document(page_content="bar")]
 
+    async def test_image_asimilarity_search(self, image_vs, image_uris):
+        results = await image_vs.asimilarity_search(image_uri=image_uris[0], k=1)
+        assert len(results) == 1
+        assert results[0].metadata["image_uri"] == image_uris[0]
+
     async def test_asimilarity_search_score(self, vs):
-        results = await vs.asimilarity_search_with_score("foo")
+        results = await vs.asimilarity_search_with_score(query="foo")
```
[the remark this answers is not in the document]
duwenxin99: updated

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: updated

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -349,6 +351,45 @@ async def aadd_documents(
...
+        metadatas: Optional[List[dict]] = None,
+        ids: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        """Embed images and add to the table."""
+        encoded_images = []
+        if metadatas is None:
+            metadatas = [{"image_uri": uri} for uri in images]
```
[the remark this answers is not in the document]
duwenxin99: If we always add uris to the users' custom metadata, what if the users do not want this field? They may find it annoying... Also it would be harder for them to go delete it than adding their own uris field.

src/langchain_google_alloydb_pg/vectorstore.py
duwenxin99: There is no more specific exception we can catch because of different embedding implementations. Yes VertexAI will soon be supporting this but there may be future image embedding implementations still not supporting list input.

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -488,6 +516,35 @@ async def asimilarity_search(
             embedding=embedding, k=k, filter=filter, **kwargs
         )
 
+    def _images_embedding_helper(self, image_uris: List[str]) -> List[List[float]]:
+        # check if `embed_image()` API is supported by the embedding service used
+        if hasattr(self.embedding_service, "embed_image"):
```
vishwarajanand, line 521: Are there examples of child `langchain_core.embeddings::Embeddings` class where `embed_image` is implemented?

duwenxin99: Currently there are:

tests/test_async_vectorstore_search.py
```diff
@@ -234,6 +280,10 @@ async def test_similarity_search(self, vs_custom):
         )
         assert results == [Document(page_content="bar")]
 
+    async def test_similarity_search_image(self, image_vs, image_uris):
+        with pytest.raises(NotImplementedError):
+            await image_vs.similarity_search_image(image_uris[0], k=1)
```
averikitsch, line 285:
```suggestion
            image_vs.similarity_search_image(image_uris[0], k=1)
```

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -294,6 +295,33 @@ async def aadd_documents(
...
+            return base64.b64encode(image_file.read()).decode("utf-8")
+
+    async def aadd_images(
+        self,
+        uris: List[str],
+        metadatas: Optional[List[dict]] = None,
+        ids: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        """Embed images and add to the table."""
```
averikitsch, line 310: Can you add the args to the doc string? We may also be able to call out the GCS support for the VertexAIEmbeddings model

averikitsch (approved):

PR #209, opened 2024-08-27
feat: Added manually tested sync methods for PGVector to AlloyDB migration

twishabansal:


src/langchain_google_alloydb_pg/engine.py
```diff
@@ -604,3 +605,268 @@ async def _aload_table_schema(
...
+        )
+        return self._fetch(
+            query=query,
+        )
+
+    def migrate_pgvector_collection(
+        self,
+        collection_name: str,
+        metadata_columns: Optional[List[Column]],
+        destination_table: Optional[str] = "",
```
averikitsch, line 787: Can we use None as the default?

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -604,3 +605,268 @@ async def _aload_table_schema(
...
+        Args:
+            collection_name (str): The collection to migrate.
+            metadata_columns (List[Column]): The metadata columns to be created to keep the data in a row-column format.
+                Optional.
+            destination_table (str): The name of the table to insert the data in.
+                Optional. defaults to collection_name.
+            use_json_metadata (bool): An option to keep the PGVector metadata as json in the AlloyDB table.
+                Default: False. Optional.
+            delete_pg_collection (bool): An option to delete the original data upon migration.
+                Default: True. Optional.
```
averikitsch, line 806: Let's default to False for safety reasons

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -604,3 +605,268 @@ async def _aload_table_schema(
...
+                Optional.
+            data (Sequence[RowMapping]): All the data (to be inserted) belonging to a pgvector collection.
+            use_json_metadata (bool): An option to keep the PGVector metadata as json in the AlloyDB table.
+                Default: False. Optional.
+        """
+        row_number = 0
+        params = {}
+
+        if use_json_metadata:
+            insert_query = f"INSERT INTO {destination_table} (LANGCHAIN_ID, CONTENT, EMBEDDING, LANGCHAIN_METADATA) VALUES"
```
averikitsch, line 666: We get these default values from the file

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -604,3 +605,268 @@ async def _aload_table_schema(
...
+                params[f"langchain_metadata_{row_number}"] = row.cmetadata
+
+                row_number += 1
+
+        elif metadata_column_names:
+            insert_query = (
+                f"INSERT INTO {destination_table} (LANGCHAIN_ID, CONTENT, EMBEDDING"
+            )
+            for column in metadata_column_names:
+                insert_query += ", " + column.upper()
```
averikitsch, line 687: is "upper" necessary?

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -604,3 +605,268 @@ async def _aload_table_schema(
...
+        Args:
+            collection_name (str): The name of the collection to get the uuid for.
+            pg_collection_table_name (str): The table name which stores the collection uuid to name mappings.
+                Default: "langchain_pg_collection". Optional.
+
+        Returns:
+            The uuid corresponding to the collection.
+        """
+        try:
+            collection_info = self._fetch(
```
averikitsch, line 626: The upcoming refactor we will be deleting fetch/execute. Any where that you use fetch or execute use the async pool directly like `async with self.engine.connect() as conn:...`. Then we will wrap the async methods with sync wrappers. For example extract_pgvector_collection could be _aextract_pgvector_collection. Then in extract_pgvector_collection we will use run_as_sync to run the _aextract_pgvector_collection methods. See https://github.com/googleapis/langchain-google-cloud-sql-pg-python/blob/c3a0f12276eef07a64ea40b20abe496496b5132c/src/langchain_google_cloud_sql_pg/engine.py#L364 for an example

src/langchain_google_alloydb_pg/engine.py
twishabansal: Changed.

src/langchain_google_alloydb_pg/engine.py
twishabansal: I see the default values hardcoded here: https://github.com/googleapis/langchain-google-alloydb-pg-python/blob/783dcad96b3cf0ae2ac11af826b71dff67a47f1f/src/langchain_google_alloydb_pg/vectorstore.py#L104

Which file can we extract them from?

src/langchain_google_alloydb_pg/engine.py
twishabansal: I'll refactor this PR once the async/sync refactor is pushed.

averikitsch:
@twishabansal  This PR can now be closed in favor of https://github.com/googleapis/langchain-google-alloydb-pg-python/pull/235, correct?

PR #228, opened 2024-09-11
feat: Made add_embeddings method public

twishabansal:
Tested Pinecone DB migration using this method: https://github.com/twishabansal/vector-store-experiments/blob/c6f264558c6dba38170000244dd85375a7e5c329/pinecone/app.py#L164

### DIFF
--- a/src/langchain_google_alloydb_pg/async_vectorstore.py
+++ b/src/langchain_google_alloydb_pg/async_vectorstore.py
@@ -217,7 +217,7 @@ class AsyncAlloyDBVectorStore(VectorStore):
     def embeddings(self) -> Embeddings:
         return self.embedding_service
 
-    async def __aadd_embeddings(
+    async def aadd_embeddings(
         self,
         texts: Iterable[str],
         embeddings: List[List[float]],
@@ -225,7 +225,7 @@ class AsyncAlloyDBVectorStore(VectorStore):
         ids: Optional[List[str]] = None,
         **kwargs: Any,
     ) -> List[str]:
-        """Add embeddings to the table."""
+        """Add data along with embeddings to the table."""
         if not ids:
             ids = [str(uuid.uuid4()) for _ in texts]
         if not metadatas:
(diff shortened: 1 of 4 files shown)


src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -217,15 +217,15 @@ async def create(
     def embeddings(self) -> Embeddings:
         return self.embedding_service
 
-    async def __aadd_embeddings(
+    async def aadd_embeddings(
```
averikitsch, line 220: Let's keep this private here since it is the pure implementation of the vector store that we don't expect users to use directly

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -766,6 +766,18 @@ def add_documents(
             "Sync methods are not implemented for AsyncAlloyDBVectorStore. Use AlloyDBVectorStore interface instead."
         )
 
+    def add_embeddings(
```
averikitsch, line 769: I do like how this makes it clear to the user this isn't support, but since it's not a part of the interface let's remove this change.

averikitsch (changes requested):

tests/test_async_vectorstore.py
```diff
@@ -205,7 +205,31 @@ async def test_adelete(self, engine, vs):
         result = await vs.adelete()
         assert result == False
 
-    ##### Custom Vector Store  #####
+    #### Custom Vector Store  #####
```
vishwarajanand, line 208: Consider removing this line

tests/test_async_vectorstore.py
```diff
@@ -205,7 +205,31 @@ async def test_adelete(self, engine, vs):
...
+            TABLE_NAME,
+            VECTOR_SIZE,
+            metadata_columns=[Column("page", "VARCHAR"), Column("source", "VARCHAR")],
+        )
+        vs = await AsyncAlloyDBVectorStore.create(
+            engine,
+            embedding_service=embeddings_service,
+            table_name=TABLE_NAME,
+            metadata_columns=["page", "source"],
+        )
```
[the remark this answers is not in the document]
vishwarajanand: Also applicable for `tests/test_vectorstore::test_add_embeddings`

averikitsch (approved):

PR #282, opened 2024-12-05
feat: Add Checkpointer base

Feromar98:
Change Log:

[ ] Fixed issues from googleapis/langchain-google-cloud-sql-pg-python#235

[ ] Create a copy of langchain's engine.py and create init_checkpoint_table methods

[ ] Finish developing checkpoint.py

### DIFF
--- a/src/langgraph_google_alloydb_pg/async_checkpoint.py
+++ b/src/langgraph_google_alloydb_pg/async_checkpoint.py
@@ -1,0 +1,109 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import asyncio
+from contextlib import asynccontextmanager
+
+import json
+from typing import List, Sequence, Any, AsyncIterator, Iterator, Optional
+
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncEngine
+
+from langchain_core.runnables import RunnableConfig
+
+from langgraph.checkpoint.base import (
+    BaseCheckpointSaver,
+    ChannelVersions,
+    Checkpoint,
+    CheckpointMetadata,
+    CheckpointTuple
+)
+from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer
+from langgraph.checkpoint.serde.types import TASKS, ChannelProtocol
+
+from langgraph.checkpoint.serde.base import SerializerProtocol
+
+MetadataInput = Optional[dict[str, Any]]
+
+from .engine import AlloyDBEngine
+
+
+class AsyncAlloyDBSaver(BaseCheckpointSaver[str]):
+    """Checkpoint stored in an AlloyDB for PostgreSQL database."""
+    
+    __create_key = object()
+    
+    def __init__(
+        self,
+        key: object,
+        pool: AsyncEngine,
+        serde: Optional[SerializerProtocol] = None
+    ) -> None:
+        super().__init__(serde=serde)
+        if key != AsyncAlloyDBSaver.__create_key:
+            raise Exception(
+                "only create class through 'create' or 'create_sync' methods"
+            )
+        self.pool = pool
+        self.serde = serde
+        
+    @classmethod
+    async def create(
+        cls,
+        engine: AlloyDBEngine,
+        serde: Optional[SerializerProtocol] = None
+    ) -> "AsyncAlloyDBSaver":
+        pass
+    
+    
+        
+    async def alist(
+        self,
+        config: Optional[RunnableConfig],
+        *,
+        filter: Optional[dict[str, Any]] = None,
+        before: Optional[RunnableConfig] = None,
+        limit: Optional[int] = None,
+    ) -> AsyncIterator[CheckpointTuple]:
+        pass
+    
+    async def aget_tuple(self):
+        pass
+    
+    async def aput(self):
+        pass
+    
+    async def aput_writes(self):
+        pass
+    
+    def list(self) -> None:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBSaver. Use AlloyDBSaver interface instead."
+        )
+        
+    def get_tuple(self) -> None:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBSaver. Use AlloyDBSaver interface instead."
+        )
+        
+    def put(self) -> None:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBSaver. Use AlloyDBSaver interface instead."
+        )
+    
+    def put_writes(self) -> None:
+        raise NotImplementedError(
+            "Sync methods are not implemented for AsyncAlloyDBSaver. Use AlloyDBSaver interface instead."
+        )
--- a/src/langgraph_google_alloydb_pg/async_checkpoint.py
+++ b/src/langgraph_google_alloydb_pg/async_checkpoint.py
@@ -18,7 +18,7 @@ import asyncpg # type: ignore
 from contextlib import asynccontextmanager
 
 import json
-from typing import List, Sequence, Any, AsyncIterator, Iterator, Optional, Dict, Tuple
+from typing import List, Sequence, Any, AsyncIterator, Iterator, Optional, cast, Dict, Tuple
 
 from sqlalchemy import text
 from sqlalchemy.ext.asyncio import AsyncEngine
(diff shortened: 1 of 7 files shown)


src/langgraph_google_alloydb_pg/engine.py
```diff
@@ -0,0 +1,521 @@
...
+        Create AlloyDB tables to save checkpoints.
+
+        Args:
+            schema_name (str): The schema name to store the checkpoint tables.
+                Default: "public".
+
+        Returns:
+            None
+        """
+        MIGRATIONS = [
```
averikitsch, line 434: I am investigating if we can do better in terms of limiting the number of tables we have and using SQL commands to pull the right data.

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,109 @@
...
+        pool: AsyncEngine,
+        serde: Optional[SerializerProtocol] = None
+    ) -> None:
+        super().__init__(serde=serde)
+        if key != AsyncAlloyDBSaver.__create_key:
+            raise Exception(
+                "only create class through 'create' or 'create_sync' methods"
+            )
+        self.pool = pool
+        self.serde = serde
```
averikitsch, line 60: this shouldn't be needed

averikitsch (commented):
How do we want to proceed with this PR? There are a few needed fixes and I need to think about the schema as well. Once the comments besides the schema are fixed, we can merge this PR and create a new one for the next bit of work, how does that sound?

src/langgraph_google_alloydb_pg/engine.py
Feromar98: Of course, I will also look for options to limit the number of tables. If you come to a conclusion, please let me know.

Feromar98:
> How do we want to proceed with this PR? There are a few needed fixes and I need to think about the schema as well. Once the comments besides the schema are fixed, we can merge this PR and create a new one for the next bit of work, how does that sound?

That sounds good to me! In the meantime, I'll create the PR for checkpointer base but for cloudSQL.
But what do you mean by the schema? That's my only doubt so far

averikitsch:
By schema I mean the table schema. The interface puts CheckpointTuples and Writes to the database and gets CheckpointTuples from the database. I think we can design a table schema where we can read and write the CheckpointTuple object and writes without needing to have 4 different tables.  Currently, I don't think we need the blobs or migration tables. That data is already stored in the checkpoint objects.

src/langgraph_google_alloydb_pg/checkpoint.py
```diff
@@ -0,0 +1,175 @@
...
+    
+
+    async def alist(
+        self,
+        config: Optional[RunnableConfig],
+        filter: Optional[dict[str, Any]] = None,
+        before: Optional[RunnableConfig] = None,
+        limit: Optional[int] = None
+    ) -> AsyncIterator[CheckpointTuple]:
+        '''List checkpoints from AlloyDB '''
```
averikitsch, line 111: Please make sure all methods have docstrings with the args and return values listed

tests/test_async_checkpoint.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2024 Google LLC
```
averikitsch, line 1: Please copy over tests from from https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/tests/test_async.py
We will need more to make sure we have enough test coverage, but copying their tests should give us confidence that the implementation is as expected.

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,352 @@
...
+            serde (SerializerProtocol): Serializer for encoding/decoding checkpoints (default: None).
+
+        Raises:
+            IndexError: If the table provided does not contain required schema.
+
+        Returns:
+            AsyncAlloyDBSaver: A newly created instance of AsyncAlloyDBSaver.
+        """
+        
+        checkpoints_table_schema = await engine._aload_table_schema("checkpoints", schema_name)
```
averikitsch, line 88: Let's use a variable for the table name "checkpoints" and "checkpoint_writes" in the engine class and import them here, so we make sure they stay consistent

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,352 @@
...
+                """
+
+        async with self.pool.connect() as conn:
+            await conn.execute(
+                text(query),
+                {
+                    "thread_id": thread_id,
+                    "checkpoint_ns": checkpoint_ns,
+                    "checkpoint_id": checkpoint_id,
+                    "parent_checkpoint_id": config.get("checkpoint_id"),
```
averikitsch, line 230: I need to check on this value

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,352 @@
...
+                text(query),
+                {
+                    "thread_id": config["configurable"]["thread_id"],
+                    "checkpoint_ns": config["configurable"]["checkpoint_ns"],
+                    "checkpoint_id": config["configurable"]["checkpoint_id"],
+                    "task_id": task_id,
+                    "idx": idx,
+                    "channel": write[0],
+                    "type": write[1],
+                    "blob": json.dumps(write[2]),
```
averikitsch, line 275: I believe this is write but note to self to check closer

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,352 @@
...
+        upsert = ""
+        async with self.pool.connect() as conn:
+            await conn.execute(
+                text(query),
+                {
+                    "thread_id": config["configurable"]["thread_id"],
+                    "checkpoint_ns": config["configurable"]["checkpoint_ns"],
+                    "checkpoint_id": config["configurable"]["checkpoint_id"],
+                    "task_id": task_id,
+                    "idx": idx,
```
averikitsch, line 272: idx is converted from writes[0] with the IDX Map, see https://github.com/langchain-ai/langgraph/blob/506539ac9df277a113423d40999518a0ad2bdb3f/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py#L299

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,352 @@
...
+                    "thread_id": thread_id,
+                    "checkpoint_ns": checkpoint_ns,
+                    "checkpoint_id": checkpoint_id,
+                    "parent_checkpoint_id": config.get("checkpoint_id"),
+                    "checkpoint": json.dumps(copy),
+                    "metadata": json.dumps(dict(metadata)),
+                    "channel": copy.pop("channel_values"),
+                    "version": new_versions,
+                    "type": next_config["configurable"]["type"],
+                    "blob": json.dumps(next_config["configurable"]["blob"]),
```
averikitsch, line 236: I don't think we need type and blob here since that is capture in the writes.

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,352 @@
...
+                text(query),
+                {
+                    "thread_id": thread_id,
+                    "checkpoint_ns": checkpoint_ns,
+                    "checkpoint_id": checkpoint_id,
+                    "parent_checkpoint_id": config.get("checkpoint_id"),
+                    "checkpoint": json.dumps(copy),
+                    "metadata": json.dumps(dict(metadata)),
+                    "channel": copy.pop("channel_values"),
+                    "version": new_versions,
```
averikitsch, line 234: I need to ask how new_versions is handled.

src/langgraph_google_alloydb_pg/engine.py
```diff
@@ -0,0 +1,530 @@
...
+            checkpoint_ns TEXT NOT NULL DEFAULT '',
+            checkpoint_id TEXT NOT NULL,
+            parent_checkpoint_id TEXT,
+            v INTEGER NOT NULL,
+            checkpoint JSONB NOT NULL,
+            metadata JSONB NOT NULL DEFAULT '{{}}',
+            channel TEXT NOT NULL,
+            version TEXT NOT NULL,
+            type TEXT,
+            blob BYTEA,
```
averikitsch, line 439: Remove type and blob. In the schema we discussed expanding the checkpoint and metadata objects into the table. This doesn't exactly match the doc or the warning check or the puts SQL statement.  We can see how just inserting the checkpoint and metadata objects as JSON into the table and see how that works.

src/langgraph_google_alloydb_pg/engine.py
```diff
@@ -454,6 +477,8 @@ async def _ainit_checkpoint_table(
         
         async with self._pool.connect() as conn:
             await conn.execute(text(create_checkpoints_table))
+            await conn.commit()
+        async with self._pool.connect() as conn:
```
averikitsch, line 481: Were these 2 lines needed to create both tables? I would expect we could call both executes then one commit like in the previous commit. If this doesn't work we should be able to execute and commit both tables within the same connection so line 481 isn't needed.

tests/test_async_checkpoint.py
```diff
@@ -36,20 +43,112 @@ async def aexecute(engine: AlloyDBEngine, query: str) -> None:
         await conn.commit()
 
 
-@pytest_asyncio.fixture
-async def async_engine():
```
averikitsch: You'll still need to set up a Saver fixture to pass into the tests

src/langgraph_google_alloydb_pg/async_checkpoint.py
```diff
@@ -161,12 +167,133 @@ async def alist(
...
+                    {
+                        "id": value["checkpoint"].get("id"),
+                        "ts": value["checkpoint"].get("ts"),
+                        "v": value["checkpoint"].get("v"),
+                        "channel_versions": value["checkpoint"].get("channel_versions"),
+                        "versions_seen": value["checkpoint"].get("versions_seen"),
+                        "pending_sends": [
+                            self.serde.loads_typed((c.decode(), b)) for c, b in value["pending_sends"] or []
+                        ],
+                        "channel_values": if not
```
averikitsch, line 265: "if not" typo or incorrect syntax

averikitsch (changes requested):

averikitsch:
Closing. We will be using smaller PRs.

PR #283, opened 2024-12-07
chore: adhere to PEP 585 and removed unused imports

vishwarajanand:
As of Python 3.9 and [PEP 585](https://peps.python.org/pep-0585/#implementation) we can use standard collections like
`list`, `dict`, `tuple` for typing instead of the  now deprecated

```python
from typing import Dict, List, Tuple ...
```

Also removing a ton of unused imports.

## Note: 
Pinned `pydantic` package version to 2.9.0 to avoid a reasoning engine failure.

### DIFF
--- a/samples/langchain_on_vertexai/requirements.txt
+++ b/samples/langchain_on_vertexai/requirements.txt
@@ -3,4 +3,6 @@ langchain-google-alloydb-pg==0.7.0
 langchain-google-vertexai==2.0.7
 google-cloud-resource-manager==1.13.0
 langchain-community==0.3.7
+# Required to fix: "PydanticUndefinedAnnotation: name 'SafetySetting' is not defined"
+# Todo: remove after upstream issue is fixed: https://github.com/langchain-ai/langchain/issues/28271
 pydantic==2.9.0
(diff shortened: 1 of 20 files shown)


samples/langchain_on_vertexai/requirements.txt
```diff
@@ -3,3 +3,6 @@ langchain-google-alloydb-pg==0.7.0
 langchain-google-vertexai==2.0.7
 google-cloud-resource-manager==1.13.0
 langchain-community==0.3.7
+# Required to fix: "PydanticUndefinedAnnotation: name 'SafetySetting' is not defined"
+# Todo: remove after upstream issue is fixed: https://github.com/langchain-ai/langchain/issues/28271
+pydantic==2.9.0
```
jackwotherspoon, line 8: is this reasoning engine fix related to this PR at all? I can't imagine it is...

If it is not then it should be split out into its own separate PR to follow open source best practices. (See [pull request best practices](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/getting-started/best-practices-for-pull-requests#write-small-prs))

This will allow a more clear history in the future.

jackwotherspoon (changes requested):

samples/langchain_on_vertexai/requirements.txt
vishwarajanand: Thanks for pointing this out. I've moved this out to another PR which should be merged first: https://github.com/googleapis/langchain-google-alloydb-pg-python/pull/284

Once this is merged, this change will go away from this PR.

averikitsch (approved):

jackwotherspoon (approved):

PR #293, opened 2024-12-17
feat: add engine_args argument to engine creation functions

kurtisvg:
Adds an "engine_args" argument to engine creation functions that can be used to specify additional parameters to the underlying pool creation.

### DIFF
--- a/tests/test_engine.py
+++ b/tests/test_engine.py
@@ -115,6 +115,11 @@ class TestEngineAsync:
             instance=db_instance,
             region=db_region,
             database=db_name,
+            engine_args={
+                # add some connection args to validate engine_args works correctly
+                "pool_size": 3,
+                "max_overflow": 2,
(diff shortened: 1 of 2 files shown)


Yuan325 (approved):

tests/test_engine.py
```diff
@@ -55,7 +55,7 @@ async def aexecute(
     query: str,
 ) -> None:
     async def run(engine, query):
-        async with engine._pool.connect() as conn:
+        async with engine._pool.pool().connect() as conn:
```
averikitsch, line 58: typo?

averikitsch (approved):

tests/test_engine.py
averikitsch, line 58:
```suggestion
        async with engine._pool.connect() as conn:
```

PR #310, opened 2025-01-10
fix: query and return only selected metadata columns

dishaprakash:
feat: Add feature to customize query


src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -534,13 +534,19 @@ async def __query_collection(
         embedding: list[float],
         k: Optional[int] = None,
         filter: Optional[str] = None,
+        columns: Optional[list[str]] = None,
```
averikitsch, line 537: Instead of changing this method signature and duplicating the code to create the column list, just do that from this method

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -673,8 +679,13 @@ async def asimilarity_search_with_score_by_vector(
         **kwargs: Any,
     ) -> list[tuple[Document, float]]:
         """Return docs and distance scores selected by vector similarity search."""
+
+        columns = self.metadata_columns + [self.content_column, self.embedding_column]
```
averikitsch, line 683: Missing id_column

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -673,8 +679,13 @@ async def asimilarity_search_with_score_by_vector(
...
+        if self.metadata_json_column:
+            columns += [self.metadata_json_column]
```
averikitsch, line 685: nit: 
```suggestion
            columns.append(self.metadata_json_column)
```

averikitsch (changes requested):

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -550,7 +563,7 @@ async def __query_collection(
             query_embedding = self.embedding_service.embed_query_inline(kwargs["query"])
         else:
             query_embedding = f"'{embedding}'"
-        stmt = f'SELECT *, {search_function}({self.embedding_column}, {query_embedding}) as distance FROM "{self.schema_name}"."{self.table_name}" {filter} ORDER BY {self.embedding_column} {operator} {query_embedding} LIMIT {k};'
+        stmt = f'SELECT {column_names}, {search_function}({self.embedding_column}, {query_embedding}) as distance FROM "{self.schema_name}"."{self.table_name}" {filter} ORDER BY {self.embedding_column} {operator} {query_embedding} LIMIT {k};'
```
vishwarajanand, line 566: Does it make sense to do `"{column_names}"` instead? 
This could help if user provides something like "table-name."

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -541,6 +541,19 @@ async def __query_collection(
...
+        columns = self.metadata_columns + [
+            self.id_column,
+            self.content_column,
+            self.embedding_column,
+        ]
+        if self.metadata_json_column:
+            columns.append(self.metadata_json_column)
+
+        if columns:
+            column_names = " ,".join(columns)
```
vishwarajanand, line 553: Does it make sense to do like this instead? This could help if user provides something like "column-name."

```suggestion
            column_names = ", ".join(f'"{col}"' for col in columns)
```

averikitsch: I like the extra type safety.

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -541,6 +541,19 @@ async def __query_collection(
...
 
+        columns = self.metadata_columns + [
+            self.id_column,
+            self.content_column,
+            self.embedding_column,
+        ]
+        if self.metadata_json_column:
+            columns.append(self.metadata_json_column)
+
+        if columns:
```
averikitsch, line 552: nit: remove if else, since columns will never be none

averikitsch (approved):
Approving, please see comments. Feel free to merge after updating.

Issue #301, opened 2025-01-06
Support multiple Index Query Options parameters


averikitsch:
The ScaNN index has 2 query option parameters: scann.num_leaves_to_search and scann.pre_reordering_num_neighbors. 

This is currently erroring with: 
```
sqlalchemy.exc.ProgrammingError: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.PostgresSyntaxError'>: syntax error at or near "."
E                   [SQL: SET LOCAL scann.num_leaves_to_search = 1, scann.pre_reordering_num_neighbors = 2;]
```

This can be solved by setting each parameter separately within the same commit of the query.

PR #312, opened 2025-01-13
fix: programming error while setting multiple query option

vishwarajanand:
Fixes https://github.com/googleapis/langchain-google-alloydb-pg-python/issues/301

Reviewer point:
1. We could simply remove the `QueryOptions.to_string()`


src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -552,9 +552,11 @@ async def __query_collection(
             query_embedding = f"'{embedding}'"
         stmt = f'SELECT *, {search_function}({self.embedding_column}, {query_embedding}) as distance FROM "{self.schema_name}"."{self.table_name}" {filter} ORDER BY {self.embedding_column} {operator} {query_embedding} LIMIT {k};'
         if self.index_query_options:
-            query_options_stmt = f"SET LOCAL {self.index_query_options.to_string()};"
             async with self.engine.connect() as conn:
-                await conn.execute(text(query_options_stmt))
+                # Set each query option individually
+                for query_option in self.index_query_options.to_string().split(","):
```
averikitsch, line 557: As discussed offline, this solution works. There is a slight performance issue where we concatenate the string then split the string. To reduce confusion with long term maintenance, I prefer a solution where the QueryOptions class has a new method like "to_parameter(self) -> List[str]", then we can set each parameter in the list

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -63,9 +63,9 @@ class ExactNearestNeighbor(BaseIndex):
 @dataclass
 class QueryOptions(ABC):
     @abstractmethod
-    def to_string(self) -> str:
-        """Convert index attributes to string."""
-        raise NotImplementedError("to_string method must be implemented by subclass")
+    def to_parameter(self) -> list[str]:
+        """Convert index attributes to map."""
```
vishwarajanand, line 67:
```suggestion
        """Convert index attributes to list of configurations."""
```

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -83,9 +83,9 @@ def index_options(self) -> str:
 class HNSWQueryOptions(QueryOptions):
     ef_search: int = 40
 
-    def to_string(self) -> str:
+    def to_parameter(self) -> list[str]:
         """Convert index attributes to string."""
```
vishwarajanand, line 87:
```suggestion
        """Convert index attributes to list of configurations."""
```

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -102,9 +102,9 @@ def index_options(self) -> str:
 class IVFFlatQueryOptions(QueryOptions):
     probes: int = 1
 
-    def to_string(self) -> str:
+    def to_parameter(self) -> list[str]:
         """Convert index attributes to string."""
```
vishwarajanand, line 106:
```suggestion
        """Convert index attributes to list of configurations."""
```

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -124,9 +124,9 @@ def index_options(self) -> str:
 class IVFQueryOptions(QueryOptions):
     probes: int = 1
 
-    def to_string(self) -> str:
+    def to_parameter(self) -> list[str]:
         """Convert index attributes to string."""
```
vishwarajanand, line 128:
```suggestion
        """Convert index attributes to list of configurations."""
```

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -147,6 +147,9 @@ class ScaNNQueryOptions(QueryOptions):
     num_leaves_to_search: int = 1
     pre_reordering_num_neighbors: int = -1
 
-    def to_string(self) -> str:
+    def to_parameter(self) -> list[str]:
         """Convert index attributes to string."""
```
vishwarajanand, line 151:
```suggestion
        """Convert index attributes to list of configurations."""
```

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -63,9 +63,9 @@ class ExactNearestNeighbor(BaseIndex):
 @dataclass
 class QueryOptions(ABC):
     @abstractmethod
-    def to_string(self) -> str:
-        """Convert index attributes to string."""
-        raise NotImplementedError("to_string method must be implemented by subclass")
+    def to_parameter(self) -> list[str]:
```
averikitsch, line 66: We will need to keep the to_string method in order for this not to be a breaking change.

averikitsch (changes requested):

averikitsch (approved):

PR #319, opened 2025-01-15
chore: update package deps

averikitsch:


requirements.txt
```diff
@@ -1,6 +1,6 @@
-google-cloud-alloydb-connector[asyncpg]==1.4.0
-google-cloud-storage>=2.18.2, <3.0.0
```
kurtisvg, line 2: Should this restriction go away?

kurtisvg (approved):

requirements.txt
averikitsch: That restriction is set in the pyproject.toml

PR #320, opened 2025-01-17
feat(langgraph-base): create checkpoint aput method + tests

Feromar98:
Change Log:

[ ] Create AsyncAlloyDBSaver class

[ ] Create async put method

[ ] Create test for class and method


src/langchain_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,222 @@
...
+        next_config: RunnableConfig = {
+            "configurable": {
+                "thread_id": thread_id,
+                "checkpoint_ns": checkpoint_ns,
+                "checkpoint_id": checkpoint["id"],
+            }
+        }
+
+        # Blobs variable
+        blobs = [
```
Feromar98, line 190: How can we use this variable? In the Postgres integration uses CHECKPOINT_BLOBS table: https://github.com/langchain-ai/langgraph/blob/76be64adcb98e086faee2825b92197191227d8ab/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py#L261C17-L261C49

UPSERT_CHECKPOINT_BLOBS_SQL: https://github.com/langchain-ai/langgraph/blob/76be64adcb98e086faee2825b92197191227d8ab/libs/checkpoint-postgres/langgraph/checkpoint/postgres/base.py#L110

PR #336, opened 2025-01-30
chore: make alloydb_table as alloydb_table_name in snippets

vishwarajanand:
This will make variable names uniform

Screenshot 2025-01-30 at 14 58 10

1. As a CONST variable in top of script (not the snippets) we are using `ALLOYDB_TABLE_NAME`
2. As a variable in the snippets, we are using `alloydb_table`


vishwarajanand:
Closing this PR because we decided not dedup the note on the docsite.

PR #356, opened 2025-02-24
chore: Add Langchain standard vectorstore tests.

dishaprakash:
Adding Langchain standard test suite for vectorstore for edge cases.

Fixes [b/375659546](https://b.corp.google.com/issues/375659546)

### DIFF
--- a/tests/test_standard_test_suite.py
+++ b/tests/test_standard_test_suite.py
@@ -1,4 +1,4 @@
-# Copyright 2024 Google LLC
+# Copyright 2025 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/tests/test_standard_test_suite.py
+++ b/tests/test_standard_test_suite.py
@@ -1,0 +1,169 @@
+import os
+import uuid
+
+import pytest
+import pytest_asyncio
+from langchain_tests.integration_tests import VectorStoreIntegrationTests
+from langchain_tests.integration_tests.vectorstores import EMBEDDING_SIZE
+from sqlalchemy import text
+
+from langchain_google_alloydb_pg import AlloyDBEngine, AlloyDBVectorStore, Column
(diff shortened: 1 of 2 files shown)


tests/test_standard_test_suite.py
```diff
@@ -0,0 +1,183 @@
...
+            async_engine,
+            embedding_service=self.get_embeddings(),
+            table_name=DEFAULT_TABLE,
+        )
+
+        yield vs
+
+    @property
+    def has_sync(self) -> bool:
+        return False  # Skip sync tests for async vector store
```
averikitsch, line 183: Let's remove the has_sync and has_async for both sets of tests so it will test both the sync and async methods  to ensure our mixed interface works.

averikitsch (approved):
I've updated this PR to be against a branch "std-tests" so we can collect all the changes before we merge into main.

averikitsch (changes requested):
Looks like the async tests are being skipped
Step #3 - "Run integration tests": tests/test_standard_test_suite.py: 24 warnings
Step #3 - "Run integration tests":   /builder/home/.local/lib/python3.9/site-packages/_pytest/python.py:148: PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.

tests/test_standard_test_suite.py
```diff
@@ -0,0 +1,183 @@
...
+            table_name=DEFAULT_TABLE_SYNC,
+        )
+        yield vs
+
+    @property
+    def has_async(self) -> bool:
+        return False  # Skip async tests for sync vector store
+
+
+class TestStandardSuiteAsync(VectorStoreIntegrationTests):
```
averikitsch, line 117: I think we need to add @pytest.mark.asyncio

tests/test_standard_test_suite.py
dishaprakash: Updated

PR #357, opened 2025-02-25
fix(asimilarity_search_image): fix bug in image search functionality

zaherweb-2:
# fix(asimilarity_search_image): fix bug in image search functionality

## Issue Description
A bug was identified in the `asimilarity_search_image()` function in the `vectorstore.py` file. The function incorrectly calls `asimilarity_search()` instead of `asimilarity_search_image()`, causing image search functionality to fail.

## Current Implementation (Incorrect)
```python
async def asimilarity_search_image(self, image_uri, k=4, filter=None, **kwargs):
    """Return docs selected by similarity search on query."""
    return await self._engine._run_as_async(
        self.__vs.asimilarity_search(image_uri, k, filter, **kwargs)
    )
```

## Corrected Implementation
```python
async def asimilarity_search_image(self, image_uri, k=4, filter=None, **kwargs):
    """Return docs selected by similarity search on image_uri."""
    return await self._engine._run_as_async(
        self.__vs.asimilarity_search_image(image_uri, k, filter, **kwargs)
    )
```

## Impact
This bug prevents image-based similarity searches from functioning correctly, as it calls the text-based search method with image parameters.

## Resolution
The fix involves changing the called method from `asimilarity_search()` to `asimilarity_search_image()` to properly utilize the image search functionality.

## Testing Recommendations
After implementing this fix, please test the function with various image inputs to ensure proper functionality.

### DIFF
--- a/src/langchain_google_alloydb_pg/vectorstore.py
+++ b/src/langchain_google_alloydb_pg/vectorstore.py
@@ -644,9 +644,9 @@ class AlloyDBVectorStore(VectorStore):
         filter: Optional[str] = None,
         **kwargs: Any,
     ) -> list[Document]:
-        """Return docs selected by similarity search on query."""
+        """Return docs selected by similarity search on image_uri."""
         return await self._engine._run_as_async(
-            self.__vs.asimilarity_search(image_uri, k, filter, **kwargs)
+            self.__vs.asimilarity_search_image(image_uri, k, filter, **kwargs)
         )
 
     # Required for (a)similarity_search_with_relevance_scores


averikitsch (approved):
Thank you!

zaherweb-2:
Thanks a lot @averikitsch!

I also have another suggestion—not a bug report—regarding the [aadd_images](https://github.com/googleapis/langchain-google-alloydb-pg-python/blob/80f888b6ee333676d4f51ff6a093cd4ca5103211/src/langchain_google_alloydb_pg/async_vectorstore.py#L341) function.

Currently, on the Alloy table where embeddings are saved, the following columns are stored:
- `langchain_id`
- `content`
- `embedding`
- `langchain_metadata`

I believe that there's no need to store the `content` (which is the image in base64) since it's already available in our cloud storage bucket. It would be great if a feature could be added to avoid duplicating this data.

PR #361, opened 2025-02-25
chore: support ids in Documents and update insert SQL query to upsert

dishaprakash:
chore: support ids in Documents and update insert SQL query to upsert

fixes [b/376065787](https://b.corp.google.com/issues/376065787)


src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -237,6 +237,8 @@ async def aadd_embeddings(
         """
         if not ids:
             ids = [str(uuid.uuid4()) for _ in texts]
+        else:
```
averikitsch, line 240: Add comment here

averikitsch (approved):

PR #382, opened 2025-03-10
feat: Enable DB-Agnostic metadata filtering in vectorstores

dishaprakash:
feat: Enable DB-Agnostic metadata filtering in vectorstores

Enabling metadata filtering using a dictionary instead of string query. 

This PR also adds 42 tests per class of vectorstore tests, to test all operators with most datatypes.


src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -550,7 +580,7 @@ async def __query_collection(
         self,
         embedding: list[float],
         k: Optional[int] = None,
-        filter: Optional[str] = None,
+        filter: Optional[dict] = None,
```
averikitsch, line 583: We should be able to support both the string and dict filters so this doesn't have to be a breaking change

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -946,6 +979,213 @@ async def aget_by_ids(self, ids: Sequence[str]) -> list[Document]:
...
+                    or_ = [self._create_filter_clause(el) for el in value]
+                    if len(or_) > 1:
+                        return f"({' OR '.join(or_)})"
+                    elif len(or_) == 1:
+                        return or_[0]
+                    else:
+                        raise ValueError(
+                            "Invalid filter condition. Expected a dictionary "
+                            "but got an empty dictionary"
+                        )
```
vishwarajanand, line 1143: These 2 can be unified

src/langchain_google_alloydb_pg/async_vectorstore.py
dishaprakash: Added the support for text based filtering too.

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -946,6 +979,199 @@ async def aget_by_ids(self, ids: Sequence[str]) -> list[Document]:
...
+                    return and_[0]
+                else:
+                    raise ValueError(
+                        "Invalid filter condition. Expected a dictionary "
+                        "but got an empty dictionary"
+                    )
+            else:
+                raise ValueError("Got an empty dictionary for filters.")
+        else:
+            raise ValueError(
```
averikitsch, line 1171: nit: it's better to return right way
```
if not isinstance(filters, dict):
  raise valueerror(...)

# the rest of the code

File: src/langchain_google_alloydb_pg/async_vectorstore.py [hunk 4]
```diff
@@ -946,6 +979,199 @@ async def aget_by_ids(self, ids: Sequence[str]) -> list[Document]:
...
+                    return f"({' AND '.join(and_)})"
+                elif len(and_) == 1:
+                    return and_[0]
+                else:
+                    raise ValueError(
+                        "Invalid filter condition. Expected a dictionary "
+                        "but got an empty dictionary"
+                    )
+            else:
+                raise ValueError("Got an empty dictionary for filters.")
```
averikitsch, line 1169: Is this what langchain-postgres is doing? I'm not sure as a user I would want this through an error if I pass in an empty dictionary

dishaprakash: Yes, this is what langchain-postgres is doing, but I could change it to not through an error

averikitsch (approved):

PR #383, opened 2025-03-10
chore: Prepare package for upstreaming langchain changes

dishaprakash:
chore: Prepare package for upstreaming langchain changes

For bug - [b/401013845](https://b.corp.google.com/issues/401013845)

This PR prepares the package for upstreaming changes to langchain-postgres.
The changes are restricted to vectorstore.
- Instead of checking for `ScaNNIndex` class instance while applying a vector index, the integration now relies on the use to specify if the `ScaNNIndex` extension needs to be created .
- Removes the check for AlloyDBEmbeddings and checks if the provided embedding class has the function `embed_query_inline`

 These changes are made so as to not have a circular import on upstreaming these changes to langchain-postgres

### DIFF
--- a/src/langchain_google_alloydb_pg/async_vectorstore.py
+++ b/src/langchain_google_alloydb_pg/async_vectorstore.py
@@ -30,7 +30,6 @@ from langchain_core.vectorstores import VectorStore, utils
 from sqlalchemy import RowMapping, text
 from sqlalchemy.ext.asyncio import AsyncEngine
 
-from .embeddings import AlloyDBEmbeddings
 from .engine import AlloyDBEngine
 from .indexes import (
     DEFAULT_DISTANCE_STRATEGY,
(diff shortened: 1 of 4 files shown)


src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -171,6 +175,20 @@ def index_options(self) -> str:
         """Set index query options for vector store initialization."""
         return f"(num_leaves = {self.num_leaves}, quantizer = {self.quantizer})"
```
averikitsch, line 177: Set enable_extension = True in the index class

tests/test_vectorstore_index.py
```diff
@@ -297,13 +297,16 @@ async def test_aapply_vector_index_ivf(self, vs):
         await vs.adrop_vector_index()
 
     async def test_aapply_alloydb_scann_index_ScaNN(self, omni_vs):
-        index = ScaNNIndex(distance_strategy=DistanceStrategy.EUCLIDEAN)
+        index = ScaNNIndex(
+            distance_strategy=DistanceStrategy.EUCLIDEAN, enable_extension=True
```
averikitsch, line 301: Remove `enable_extension=True` from here and have the class handle the default

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -826,11 +829,9 @@ async def aapply_vector_index(
...
 
         # Create `alloydb_scann` extension when a `ScaNN` index is applied
-        if isinstance(index, ScaNNIndex):
-            async with self.engine.connect() as conn:
-                await conn.execute(text("CREATE EXTENSION IF NOT EXISTS alloydb_scann"))
-                await conn.commit()
-            function = index.distance_strategy.scann_index_function
+        if index.enable_extension:
+            if callable(getattr(index, "create_extension", None)):
+                function = await index.create_extension(self.engine)  # type: ignore
```
averikitsch, line 834: I want to keep all engine connection code in the async_vectorstore vs the index classes. I think we can use extension_name attribute as a flag instead of enable_extension. In the baseindex we would set extension_name = None and in the scann class extension_name= alloydb_scann
```suggestion
        if index.extension_name:
            async with self.engine.connect() as conn:
                await conn.execute(text("CREATE EXTENSION IF NOT EXISTS {index.extension_name}"))
                await conn.commit()
```

src/langchain_google_alloydb_pg/indexes.py
averikitsch, line 1: The design doc missed that we also need to remove scann_index_function and support a general way to change index_function. I need to think about this longer.

src/langchain_google_alloydb_pg/async_vectorstore.py
dishaprakash: Removed the usage of enable_extension and added extension_name

PR #391, opened 2025-03-13
feat(langgraph): add checkpointer interface

averikitsch:

### DIFF
--- a/src/langchain_google_alloydb_pg/async_checkpoint.py
+++ b/src/langchain_google_alloydb_pg/async_checkpoint.py
@@ -404,7 +404,7 @@ class AsyncAlloyDBSaver(BaseCheckpointSaver[str]):
                     break
                 value = row._mapping
                 metadata: CheckpointMetadata = (
-                    self.jsonplus_serde.loads(value["metadata"])
+                    dict(self.jsonplus_serde.loads(value["metadata"]))
                     if value["metadata"] is not None
                     else {}
                 )
@@ -501,7 +501,7 @@ class AsyncAlloyDBSaver(BaseCheckpointSaver[str]):
                 },
                 checkpoint=self.serde.loads_typed((value["type"], value["checkpoint"])),
                 metadata=(
-                    self.jsonplus_serde.loads(value["metadata"])
+                    dict(self.jsonplus_serde.loads(value["metadata"]))
                     if value["metadata"] is not None
                     else {}
                 ),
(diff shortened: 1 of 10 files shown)


src/langchain_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,595 @@
...
+                "\n    metadata JSONB NOT NULL"
+                "\n);"
+            )
+
+        checkpoint_writes_table_schema = await engine._aload_table_schema(
+            f"{table_name}_writes", schema_name
+        )
+        checkpoint_writes_column_names = checkpoint_writes_table_schema.columns.keys()
+
+        checkpoint_writes_columns = [
```
kurtisvg, line 124: nit: also candidate for static or global

kurtisvg: also consider https://docs.python.org/3/library/typing.html#typing.Final

src/langchain_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,595 @@
...
+            RunnableConfig: Updated configuration after storing the checkpoint.
+        """
+        configurable = config["configurable"].copy()
+        thread_id = configurable.pop("thread_id")
+        checkpoint_ns = configurable.pop("checkpoint_ns")
+        checkpoint_id = configurable.pop(
+            "checkpoint_id", configurable.pop("thread_ts", None)
+        )
+
+        next_config: RunnableConfig = {
```
kurtisvg, line 264: I don't follow this logic -- why is there a copy/pop? configureable doesn't seem to be used anywhere

src/langchain_google_alloydb_pg/async_checkpoint.py
```diff
@@ -0,0 +1,595 @@
...
+    ) -> None:
+        super().__init__(serde=serde)
+        if key != AsyncAlloyDBSaver.__create_key:
+            raise Exception(
+                "only create class through 'create' or 'create_sync' methods"
+            )
+        self.pool = pool
+        self.table_name = table_name
+        self.table_name_writes = f"{table_name}_writes"
+        self.schema_name = schema_name
```
kurtisvg, line 62: Should these be publicly accessible fields

kurtisvg (approved):

src/langchain_google_alloydb_pg/async_checkpoint.py
averikitsch: We could make them private; however we are currently following the precedent pattern.

PR #393, opened 2025-03-13
fix(engine): loop error on close

averikitsch:

### DIFF
--- a/src/langchain_google_alloydb_pg/engine.py
+++ b/src/langchain_google_alloydb_pg/engine.py
@@ -428,7 +428,8 @@ class AlloyDBEngine:
 
     async def close(self) -> None:
         """Dispose of connection pool"""
-        await self._run_as_async(self._connector.close())
+        if self._connector:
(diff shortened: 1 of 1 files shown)


vishwarajanand (approved):
This seems to be required for other integrations as well.

averikitsch:
> Approved to unblock.
> 
> This seems to be required for other integrations as well.
> 
> Although this fixes the errors for `attached to a different loop`, I still see errors like this when I explicitly call `await engine.close()`:
> 
> ```
> Unclosed client session
> client_session: <aiohttp.client.ClientSession object at 0x7f90ed73e110>
> Unclosed connector
> connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x7f90e41eac10>, 30576.20452686), (<aiohttp.client_proto.ResponseHandler object at 0x7f90e41eab30>, 30576.352191583)]']
> connector: <aiohttp.connector.TCPConnector object at 0x7f90ded43650>
> ```

Yes, we will. I've created b/403606267 to track this other work.

PR #407, opened 2025-04-21
feat(alloydb_vectorstore): Implement batch insertion in aadd_embeddings

YoussefSaad1:
**Description:**

Replaces the previous implementation of `aadd_embeddings` which performed one INSERT operation per embedding.

The new implementation constructs a single INSERT statement with multiple VALUES clauses, allowing for significantly faster insertion of multiple embeddings by reducing database roundtrips and transaction overhead.

**Problem:**
The previous implementation iterated through the provided texts and embeddings, executing a separate `INSERT ... ON CONFLICT DO UPDATE` SQL statement for each item. This resulted in numerous database roundtrips and transaction overhead, leading to poor performance when adding a large number of embeddings.

**Solution:**
The updated code now constructs a single `INSERT` statement containing multiple `VALUES` clauses, one for each text/embedding pair. It dynamically builds the necessary parameterized query string and the corresponding parameter dictionary. This entire batch is then executed in a single database call and committed once.

**Benefits:**
- Significantly improves the performance of adding multiple embeddings.
- Reduces network latency between the application and the database.
- Decreases the load on the database by minimizing individual query parsing and transaction handling.

### DIFF
--- a/src/langchain_google_alloydb_pg/async_vectorstore.py
+++ b/src/langchain_google_alloydb_pg/async_vectorstore.py
@@ -268,66 +268,66 @@ class AsyncAlloyDBVectorStore(VectorStore):
         if not ids:
             ids = [str(uuid.uuid4()) for _ in texts]
         else:
-            # This is done to fill in any missing ids
             ids = [id if id is not None else str(uuid.uuid4()) for id in ids]
         if not metadatas:
             metadatas = [{} for _ in texts]
-        # Insert embeddings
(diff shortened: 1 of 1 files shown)


averikitsch:
Thank you @YoussefSaad1 for your contribution! We are in the middle of refactoring this integration to wrap the PGVectorStore, since we have upstreamed this package to langchain-postgres under PGVectorStore, see https://github.com/langchain-ai/langchain-postgres/blob/main/langchain_postgres/v2. We would love to accept your contribution in that repo. Let me know if you have any questions!

PR #414, opened 2025-05-06
feat: Added Hybrid Search Config and Tests [1/N]

vishwarajanand:
Adds configs and tests for Hybrid Search based on TSV + Vector Search


vishwarajanand:
Migrated over to https://github.com/langchain-ai/langchain-postgres/pull/211

PR #419, opened 2025-05-08
tests: upgrade embedding model in tests

vishwarajanand:
Upgrade embedding models and associated scores in tests.

Since, latest models are not pre-registered, adding a call to specifically call `CREATE_MODEL` endpoint.

### DIFF
--- a/tests/test_embeddings.py
+++ b/tests/test_embeddings.py
@@ -19,7 +19,11 @@ import pytest
 import pytest_asyncio
 from langchain_core.documents import Document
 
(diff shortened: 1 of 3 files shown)


dishaprakash (approved):

dishaprakash:
@vishwarajanand Has there been any recent failure due to the score threshold being higher?

averikitsch (approved):

vishwarajanand:
> @vishwarajanand Has there been any recent failure due to the score threshold being higher?

I dont know of any! i think the scores changed because of the change in embedding models.

PR #428, opened 2025-05-26
feat(aadd_images): add support for store_uri_only parameter

zaherweb-2:
# ✨ feat(aadd_images): add support for store_uri_only parameter

## ⚠️ Issue
The `content` column currently stores the image in base64 format. Since these images are often already available in cloud storage (e.g., a GCS bucket), storing them again in the AlloyDB table leads to data duplication.

## 💡 Proposed Improvement
**Feature Addition:** Introduce an optional `store_uri_only` parameter to the image adding methods (`aadd_images`, `add_images`). This flag allows users to bypass storing the image data (base64) in the `content` column of the AlloyDB table.
**Data Source:** When this option is used, the system will rely on the existing images in their original storage location (e.g., cloud storage bucket) to avoid redundancy, storing only the URI in the `content` column. The image embedding will still be generated from the actual image content fetched from the URI.

## 🎉 Benefits
-   **Reduced Storage Overhead 💾:** Eliminates duplicate storage of image data, saving space in the AlloyDB table.
-   **Improved Efficiency 🚀:** Streamlines data management by avoiding unnecessary data transfers and encoding if the base64 content is not needed in the table.
-   **Simplified Data Model 📄:** Keeps the AlloyDB table focused on storing embeddings and essential metadata, with image URIs acting as pointers to the original image data.

This pull request introduces the `store_uri_only` parameter to achieve these benefits, providing users with more flexibility in how they manage image data within the AlloyDB vector store.

## 📝 Summary of Changes

The `store_uri_only` parameter, when set to `True`, instructs the vector store to save the provided image URI directly into the content column. If `False` (the default), the image content is fetched from the URI, base64 encoded, and then stored in the content column.

Regardless of this flag's value:
- The image embedding is always generated from the actual visual content of the image (fetched from the URI).
- The original image URI is included in the metadata (specifically in the `image_uri` key within the metadata JSON blob or a dedicated column if configured) to ensure it's always accessible.

This enhancement allows for more storage-efficient solutions when the original image data is not required to be stored directly alongside the embedding, while still enabling powerful multi-modal search capabilities.

## 📂 Modified Files

1.  **`src/langchain_google_alloydb_pg/async_vectorstore.py`**:
    *   Modified the `AsyncAlloyDBVectorStore.aadd_images` method:
        *   Added the `store_uri_only: bool = False` parameter to the method signature and updated its docstring.
        *   Implemented conditional logic to determine whether to store the image URI or its base64 encoded content in the `content_column` based on the `store_uri_only` flag.
        *   Ensured that the `image_uri` is present in the `metadatas` if `store_uri_only` is true.

2.  **`src/langchain_google_alloydb_pg/vectorstore.py`**:
    *   Modified the `AlloyDBVectorStore.aadd_images` method (asynchronous wrapper):
        *   Added the `store_uri_only: bool = False` parameter and updated its docstring.
        *   Passed the `store_uri_only` flag to the underlying call to `self.__vs.aadd_images`.
    *   Modified the `AlloyDBVectorStore.add_images` method (synchronous public method):
        *   Added the `store_uri_only: bool = False` parameter and updated its docstring.
        *   Passed the `store_uri_only` flag to the underlying call.

3.  **`tests/test_async_vectorstore.py`**:
    *   Added a new test method `test_aadd_images_store_uri_only`.
    *   This test verifies that when `store_uri_only=True`:
        *   The image URI is correctly stored in the content column.
        *   The image embedding is generated from the image content, not the URI string.
        *   Metadata, including `image_uri`, is correctly populated.

4.  **`tests/test_vectorstore.py`**:
    *   Added a new test method `test_aadd_images_store_uri_only` for the `AlloyDBVectorStore`'s async `aadd_images` method.
    *   Added a new test method `test_add_images_store_uri_only` for the `AlloyDBVectorStore`'s sync `add_images` method.
    *   These tests ensure that the `store_uri_only=True` functionality works as expected for both synchronous and asynchronous invocation paths, verifying correct URI storage in the content column and proper embedding generation.


zaherweb-2:
Hello @vishwarajanand  , should I do any changes from my end . 
If so pls let me know what should I do , like adding black check , and langchain version ? 
Thanks in advance.

tests/test_vectorstore.py
```diff
@@ -359,6 +359,51 @@ async def test_aadd_images(self, engine_sync, image_uris):
...
+                result_row[vs._AlloyDBVectorStore__vs.content_column] == image_uris[i]
+            )
+            uri_embedding = embeddings_service.embed_query(image_uris[i])
+            actual_embedding = image_embedding_service.embed_query(image_uris[i])
+            assert result_row[vs._AlloyDBVectorStore__vs.embedding_column] != str(
+                uri_embedding
+            )
+            assert result_row[vs._AlloyDBVectorStore__vs.embedding_column] == str(
+                actual_embedding
+            )
```
vishwarajanand, line 397: Can you convert the embeddings to embeddings string like this instead of directly parsing it into one:

`str([float(dimension) for dimension in embedding])`

tests/test_vectorstore.py
```diff
@@ -405,6 +450,48 @@ async def test_add_images(self, engine_sync, image_uris):
...
```
vishwarajanand, line 485: Same here

tests/test_vectorstore.py
vishwarajanand: Added in [bed7951](https://github.com/googleapis/langchain-google-alloydb-pg-python/pull/428/commits/bed7951b055d68802b252acd34550397b556ea32)

averikitsch (approved):

PR #435, opened 2025-07-24
refactor!: Refactor AlloyDBVectorStore to depend on PGVectorstore

dishaprakash:
refactor!: Refactor AlloyDBVectorStore to depend on PGVectorstore


tests/test_async_vectorstore_from_methods.py
averikitsch, line 1: Can we keep some of these tests to ensure we are not breaking the interface?

PR #458, opened 2025-08-20
chore: Disable numpy updates for python versions <= 3.10

dishaprakash:


dishaprakash:
> Using the dry run `LOG_LEVEL=debug npx renovate --platform=local`, I see that the numpy versions may still be getting updated.

Thanks Averi, I've updated the config after trying the dry run locally now. Please take a look!

PR #517, opened 2025-12-02
feat: Disable support for python 3.9 and enable support for python3.13

dishaprakash:
chore!: Disable support for python 3.9 and enable support for python3.13


dishaprakash:
Once approved, the Py3.9 test will be removed and a Py3.13 trigger will be added before merging.

PR #531, opened 2025-12-16
feat: Update Langgraph deps to new major

dishaprakash:
This PR updates Langgraph to v1 and Langgraph-checkpoint to v3


pyproject.toml
```diff
@@ -40,12 +40,12 @@ Changelog = "https://github.com/googleapis/langchain-google-alloydb-pg-python/bl
 
 [project.optional-dependencies]
 langgraph = [
-    "langgraph-checkpoint>=2.0.9, <3.0.0"
+    "langgraph-checkpoint>=3.0.0, <3.1.0"
```
averikitsch: We should be able to keep the lower bound the same

dishaprakash: We're increasing the lower bound because of this security vulnerability that was issued - https://github.com/langchain-ai/langgraph/security/advisories/GHSA-wwqv-p2pp-99h5

PR #588, opened 2026-04-13
fix: avoid mutable default argument for metadata_columns

AbedSalekin:
Fixes a Python mutable default argument issue in metadata_columns.

Replaces default [] with None and initializes safely inside functions:
if metadata_columns is None:
    metadata_columns = []

This prevents unintended shared state across function calls


dishaprakash (approved):

dishaprakash:
@AbedSalekin Thank you for raising this PR!

averikitsch (approved):

PR #621, opened 2026-07-15
feat: Vector Index & Columnar Engine Optimizations

madamsetty-pavan:
Part 1 of 3 for AlloyDB AI Features Epic. Contains Vector Optimizations including ScaNN and Columnar Engine.


src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +146,93 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+            await conn.execute(text(query))
+            await conn.commit()
+
+    async def adefine_vector_assist_spec(self) -> list[dict]:
+        """Asynchronously define a Vector Assist spec for the current table."""
+        query = "SELECT * FROM vector_assist.define_spec(table_name => :table_name, vector_column_name => :embedding_column)"
+        params = {"table_name": self.table_name, "embedding_column": self.embedding_column}
+        async with self.engine.connect() as conn:
+            result = await conn.execute(text(query), params)
+            return [dict(row._mapping) for row in result.fetchall()]
```
dishaprakash: nit: Can we use the public method mappings()?

```suggestion
        async with self.engine.connect() as conn:
            result = await conn.execute(text(query), params)
            result_map = result.mappings()
            results = result_map.fetchall()
```

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +146,93 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+
+    async def aapply_vector_assist_spec(self) -> list[dict]:
+        """Asynchronously apply the Vector Assist spec for the current table."""
+        query = "SELECT * FROM vector_assist.apply_spec(table_name => :table_name, column_name => :embedding_column)"
+        params = {"table_name": self.table_name, "embedding_column": self.embedding_column}
+        async with self.engine.connect() as conn:
+            result = await conn.execute(text(query), params)
+            return [dict(row._mapping) for row in result.fetchall()]
```
dishaprakash: nit: Same as above

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +146,93 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+            return []
+        
+        spec_id = specs[0].get("vector_spec_id")
+        if not spec_id:
+            return []
+            
+        query = "SELECT * FROM vector_assist.get_recommendations(:spec_id)"
+        async with self.engine.connect() as conn:
+            result = await conn.execute(text(query), {"spec_id": spec_id})
+            return [dict(row._mapping) for row in result.fetchall()]
```
dishaprakash: nit: Same as above

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -621,6 +621,77 @@ def init_checkpoint_table(
...
+    async def aforecast(
+        self,
+        model_id: str,
+        source_table: str,
+        timestamp_col: str,
+        data_col: str,
+        horizon: int,
+        source_query: Optional[str] = None,
+        conf_level: Optional[float] = None,
+    ) -> list[dict]:
```
dishaprakash, line 682: Can we seperate the implementation into a private asynchronous method and use the `_run_as_async` on the exposed oublic function to call it?

Similar to the `_ainit_checkpoint_table` and its public APIs `ainit_checkpoint_table` and `init_checkpoint_table`

src/langchain_google_alloydb_pg/vectorstore.py
```diff
@@ -167,6 +167,48 @@ def create_sync(
...
+
+        Args:
+            model_id (str): The model id used for generating embeddings.
+            table_name (str): Name of the table.
+            content_column (str): Name of the content column. Defaults to "content".
+            embedding_column (str): Name of the embedding column. Defaults to "embedding".
+        """
+        await self._engine._run_as_async(
+            self._PGVectorStore__vs.ainitialize_auto_vector_embeddings(  # type: ignore
+                model_id, table_name, content_column, embedding_column
```
dishaprakash: Should the table name be referenced internally?

The async implementation above does not have a table_name argument

tests/test_async_vectorstore.py
```diff
@@ -473,3 +473,51 @@ async def test_create_vectorstore_with_init(self, engine):
                 embedding_column="myembedding",
                 metadata_columns=["random_column"],  # invalid metadata column
             )
+
+    async def test_aenable_columnar_engine(self, vs):
+        """Test enabling the columnar engine triggers the appropriate async method on the underlying store."""
+        from unittest.mock import AsyncMock, patch
```
dishaprakash: nit: We can import these modules once at the top of the file?

dishaprakash (commented):
Can we also add functional tests that use the live database?
We can use fixtures to setup and teardown to bring them to original states.

src/langchain_google_alloydb_pg/async_vectorstore.py
madamsetty-pavan: Done. Updated all occurrences across `engine.py` and `async_vectorstore.py` to use `result.mappings().fetchall()` instead of the private `row._mapping`.

src/langchain_google_alloydb_pg/async_vectorstore.py
madamsetty-pavan: Done. Updated all occurrences across `engine.py` and `async_vectorstore.py` to use `result.mappings().fetchall()` instead of the private `row._mapping`.

src/langchain_google_alloydb_pg/async_vectorstore.py
madamsetty-pavan: Done

src/langchain_google_alloydb_pg/engine.py
madamsetty-pavan: Done. Extracted the core query execution into `_aforecast` and refactored both `aforecast` (via `await self._run_as_async`) and `forecast` (via `self._run_as_sync`) to delegate to it, following the exact `_ainit_checkpoint_table` pattern.

src/langchain_google_alloydb_pg/vectorstore.py
madamsetty-pavan: Removed `table_name` from the method signature in both `vectorstore.py` and `async_vectorstore.py`. Both now reference `self.table_name` internally.

tests/test_async_vectorstore.py
madamsetty-pavan: Done. Moved all `unittest.mock` imports (`AsyncMock`, `MagicMock`, `patch`) to the top level of the test files and formatted with `isort`.

madamsetty-pavan:
Done. Added live database functional tests using setup and teardown fixtures:
- `test_aapply_alloydb_scann_index_auto_mode` in `test_vectorstore_index.py` / `test_async_vectorstore_index.py` to create, validate, and drop ScaNN AUTO mode indexes on live AlloyDB instances.
- `test_live_columnar_engine`, `test_live_auto_columnarization`, and `test_live_vector_assist` in `test_vectorstore.py` / `test_async_vectorstore.py` to test columnar engine and vector assist against live AlloyDB instances.

twishabansal:
Please fix the integration tests

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -84,20 +88,28 @@ def get_index_function(self) -> str:
...
     def to_parameter(self) -> list[str]:
         """Convert index attributes to list of configurations."""
-        return [
-            f"scann.num_leaves_to_search = {self.num_leaves_to_search}",
-            f"scann.pre_reordering_num_neighbors = {self.pre_reordering_num_neighbors}",
-        ]
+        params = []
+        if self.pct_leaves_to_search is not None:
+            params.append(f"scann.pct_leaves_to_search = {self.pct_leaves_to_search}")
+        if self.num_leaves_to_search is not None:
```
twishabansal: ScaNNQueryOptions() with no arguments used to emit scann.num_leaves_to_search = 1. After this change it emits scann.pct_leaves_to_search = 1 instead. T

hose aren't equivalent: one is an absolute leaf count, the other a proportion: so existing users get different recall/latency on upgrade with no error or warning. Could we keep num_leaves_to_search: int = 1 as the default and emit pct_leaves_to_search only when the user sets it? If the default change is intentional, it should be called out as a breaking change in the changelog.

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -63,14 +64,17 @@ def to_string(self) -> str:
...
+    mode: Optional[str] = None
+    num_leaves: Optional[int] = 5
     quantizer: str = field(
         default="sq8", init=False
     )  # Disable `quantizer` initialization currently only supports the value "sq8"
     extension_name: str = "alloydb_scann"
 
     def index_options(self) -> str:
         """Set index query options for vector store initialization."""
+        if self.mode and self.mode.upper() == "AUTO":
```
twishabansal: nit: Should we add a validation here? If mode is not auto, it should emit a warning

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +150,108 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+        """
+        content_col = content_column or self.content_column
+        embedding_col = embedding_column or self.embedding_column
+        query = "CALL ai.initialize_embeddings(:model_id, :table_name, :content_column, :embedding_column)"
+        async with self.engine.connect() as conn:
+            await conn.execute(
+                text(query),
+                {
+                    "model_id": model_id,
+                    "table_name": self.table_name,
```
twishabansal: Should we be adding a schema_name arg here as well? Vectorstore does support schema_name

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +150,108 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+            result_map = result.mappings()
+            results = result_map.fetchall()
+            return [dict(row) for row in results]
+
+    async def aget_vector_assist_recommendations(self) -> list[dict]:
+        """Asynchronously get Vector Assist recommendations for the current table."""
+        # First we need to get the spec ID for the current table
+        specs = await self.adefine_vector_assist_spec()
+        if not specs:
+            return []
```
twishabansal, line 260: Should we have better error handling and logging here?

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +150,108 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+
+        spec_id = specs[0].get("vector_spec_id")
+        if not spec_id:
+            return []
```
twishabansal, line 268: Should we have better error handling and logging here?

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -63,14 +64,17 @@ def to_string(self) -> str:
...
+            return f"(mode = 'AUTO')"
```
twishabansal:
```suggestion
            return "(mode = 'AUTO')"
```

src/langchain_google_alloydb_pg/engine.py
```diff
@@ -621,6 +621,114 @@ def init_checkpoint_table(
...
+            data_col => :data_col,
+            timestamp_col => :timestamp_col,
+            horizon => :horizon,
+            conf_level => :conf_level
+        )
+        """
+        params = {
+            "model_id": model_id,
+            "source_table": source_table,
+            "source_query": source_query,
```
twishabansal: Do we need a null check for source_query and conf_level?

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +150,108 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+        content_col = content_column or self.content_column
+        embedding_col = embedding_column or self.embedding_column
+        query = "CALL ai.initialize_embeddings(:model_id, :table_name, :content_column, :embedding_column)"
+        async with self.engine.connect() as conn:
+            await conn.execute(
+                text(query),
+                {
+                    "model_id": model_id,
+                    "table_name": self.table_name,
+                    "content_column": content_col,
```
twishabansal, line 196: Should we add a null check for content_column and embedding_column?

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +150,108 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+    async def adefine_vector_assist_spec(self) -> list[dict]:
+        """Asynchronously define a Vector Assist spec for the current table."""
+        query = "SELECT * FROM vector_assist.define_spec(table_name => :table_name, vector_column_name => :embedding_column)"
+        params = {
+            "table_name": self.table_name,
+            "embedding_column": self.embedding_column,
+        }
+        async with self.engine.connect() as conn:
+            result = await conn.execute(text(query), params)
+            result_map = result.mappings()
```
twishabansal: nit: We can just use 

`return [dict(row) for row in result.mappings()]` instead of the next three lines.
This can be cleaned up at multiple places in this PR

src/langchain_google_alloydb_pg/indexes.py
```diff
@@ -84,20 +88,28 @@ def get_index_function(self) -> str:
...
         """Convert index attributes to list of configurations."""
-        return [
-            f"scann.num_leaves_to_search = {self.num_leaves_to_search}",
-            f"scann.pre_reordering_num_neighbors = {self.pre_reordering_num_neighbors}",
-        ]
+        params = []
+        if self.pct_leaves_to_search is not None:
+            params.append(f"scann.pct_leaves_to_search = {self.pct_leaves_to_search}")
+        if self.num_leaves_to_search is not None:
+            params.append(f"scann.num_leaves_to_search = {self.num_leaves_to_search}")
```
twishabansal, line 129: What happens if `pct_leaves_to_search` and `num_leaves_to_search` both are set? Should we document this behaviour or link to somewhere?

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -146,6 +150,108 @@ async def set_maintenance_work_mem(self, num_leaves: int, vector_size: int) -> N
...
+
+    async def aget_vector_assist_recommendations(self) -> list[dict]:
+        """Asynchronously get Vector Assist recommendations for the current table."""
+        # First we need to get the spec ID for the current table
+        specs = await self.adefine_vector_assist_spec()
+        if not specs:
+            return []
+
+        spec_id = specs[0].get("vector_spec_id")
+        if not spec_id:
```
twishabansal: Should `spec_id=0` be considered valid? It would also be true for `if not spec_id`

twishabansal (changes requested):

tests/test_async_vectorstore.py
```diff
@@ -473,3 +478,161 @@ async def test_create_vectorstore_with_init(self, engine):
...
+        vs.embedding_column = "embedding"
+        return vs
+
+    async def test_aenable_columnar_engine(self, vs):
+        """Test enabling the columnar engine executes queries on engine."""
+        with patch.object(vs.engine, "connect") as mock_connect:
+            mock_conn = AsyncMock()
+            mock_connect.return_value.__aenter__.return_value = mock_conn
+            await vs.aenable_columnar_engine(["content"])
+            assert mock_conn.execute.called
```
twishabansal: Ideally we should assert on some data not just that something is executed on the table? Maybe the generated sql and the results. Is there some limitations to doing that here?

Same goes for multiple places at which this assertion is present.

tests/test_async_vectorstore.py
```diff
@@ -473,3 +478,161 @@ async def test_create_vectorstore_with_init(self, engine):
...
+        await vs.aenable_columnar_engine()
+
+    async def test_live_auto_columnarization(self, vs):
+        """Test triggering auto columnarization recommendations against live AlloyDB instance."""
+        await vs.aenable_auto_columnarization()
+
+    async def test_live_vector_assist(self, vs):
+        """Test vector assist spec definition, application, and recommendations against live AlloyDB instance."""
+        specs = await vs.adefine_vector_assist_spec()
+        assert isinstance(specs, list)
```
twishabansal: Should we have stronger assertions based on the content or structure instead? Same goes for other tests in this PR

PR #629, opened 2026-08-12
feat: add Columnar Engine, Auto-Columnarization, and Vector Assist op…

madamsetty-pavan:
feat: add Columnar Engine, Auto-Columnarization, and Vector Assist optimizations

- Add enable_columnar_engine and aenable_columnar_engine
- Add enable_auto_columnarization and aenable_auto_columnarization
- Add Vector Assist spec definition, application, and recommendations
- Add initialize_auto_vector_embeddings for automated embedding generation
- Add comprehensive sync and async unit tests with detailed comments

### DIFF
--- a/src/langchain_google_alloydb_pg/async_vectorstore.py
+++ b/src/langchain_google_alloydb_pg/async_vectorstore.py
@@ -30,6 +30,11 @@ from sqlalchemy import text
 logger = logging.getLogger(__name__)
 
 
+def _quote_ident(ident: str) -> str:
+    """Quote a PostgreSQL identifier to prevent SQL injection and syntax errors."""
+    return '"' + ident.replace('"', '""') + '"'
+
+
 class AsyncAlloyDBVectorStore(AsyncPGVectorStore):
     """Google AlloyDB Vector Store class"""
 
@@ -153,10 +158,10 @@ class AsyncAlloyDBVectorStore(AsyncPGVectorStore):
         index_memory_required = max(
             10, round(50 * num_leaves * vector_size * 4 / 1024 / 1024) + buffer
         )
-        query = f"SET maintenance_work_mem TO '{index_memory_required} MB';"
-        async with self.engine.connect() as conn:
-            await conn.execute(text(query))
-            await conn.commit()
+        async with self._pool_engine.begin() as conn:
+            await conn.execute(
+                text(f"SET LOCAL maintenance_work_mem TO '{index_memory_required} MB';")
+            )
 
     set_maintenance_work_mem = aset_maintenance_work_mem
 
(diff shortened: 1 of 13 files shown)


src/langchain_google_alloydb_pg/engine.py
```diff
@@ -367,6 +370,82 @@ def from_engine_args(
...
+        schema_name: str = "public",
+    ) -> None:
+        """Asynchronously drop a table from the database.
+
+        Args:
+            table_name (str): The name of the table to drop.
+            schema_name (str): The schema name of the table. Default: "public".
+        """
+        await self._run_as_async(
+            self._adrop_table(table_name=table_name, schema_name=schema_name)
```
zhou0ricky: Is _adrop_table implemented anywhere?

src/langchain_google_alloydb_pg/async_vectorstore.py
```diff
@@ -134,17 +147,380 @@ async def asimilarity_search_image(
...
+        # Note: CREATE EXTENSION is omitted here as it requires SUPERUSER privileges.
+        # Extensions should be created during database setup by an administrator.
+
+        function = index.get_index_function()
+
+        filter = f"WHERE ({index.partial_indexes})" if index.partial_indexes else ""
+        params = "WITH " + index.index_options()
+        if name is None:
+            if index.name is None:
+                index.name = self.table_name + DEFAULT_INDEX_NAME_SUFFIX
```
zhou0ricky: Not sure if index.name is ever used elsewhere, but just confirming that index.name mutation is intentional.

zhou0ricky (approved):
Added a few comments, otherwise should be good.

src/langchain_google_alloydb_pg/engine.py
madamsetty-pavan: Good catch! `_adrop_table`, `adrop_table`, and `drop_table` are all already implemented in the base class PGEngine (langchain_postgres.v2.engine). Re-declaring `adrop_table` and `drop_table` here was redundant and made `_adrop_table` look like a missing private method in this repo's diff. 

Removed the duplicate declarations so AlloyDBEngine inherits cleanly from PGEngine (matching `init_vectorstore_table`), and added unit test coverage in tests/test_engine.py

src/langchain_google_alloydb_pg/async_vectorstore.py
madamsetty-pavan: Fixed! We now compute the default index name locally as` name = index.name or (self.table_name + DEFAULT_INDEX_NAME_SUFFIX)` without mutating `index.name` on the caller's index object, preserving object immutability. Unit tests added to verify this behavior.

PR #640, opened 2026-09-10
feat(embeddings): add templatized inline embedding support and tests

dishaprakash:
feat(embeddings): add templatized inline embedding support and tests
- Add embed_query_inline_template to AlloyDBEmbeddings returning parameterized SQL template
- Add unit test for embed_query_inline_template in test_embeddings.py
- Add regression tests with temp table assertions in test_vectorstore_embeddings.py

### DIFF
--- a/src/langchain_google_alloydb_pg/embeddings.py
+++ b/src/langchain_google_alloydb_pg/embeddings.py
@@ -128,6 +128,10 @@ class AlloyDBEmbeddings(Embeddings):
             "Embedding functions are not implemented. Use VertexAIEmbeddings interface instead."
         )
 
+    def embed_query_inline_template(self, param_name: str = ":content") -> str:
+        clean_model_id = self.model_id.replace("'", "''")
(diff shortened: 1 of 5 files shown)


src/langchain_google_alloydb_pg/embeddings.py
```diff
@@ -128,6 +128,10 @@ async def aembed_documents(self, texts: list[str]) -> list[list[float]]:
             "Embedding functions are not implemented. Use VertexAIEmbeddings interface instead."
         )
 
+    def embed_query_inline_template(self, param_name: str = ":content") -> str:
+        clean_model_id = self.model_id.replace("'", "''")
```
averikitsch, line 132: We should clean the param_name as well.

dishaprakash: Done! Added tests as well

averikitsch (approved):

